PR #8, opened 2015-11-02
Add ability to compute the next maximum ID by hand in MongoDB

zedxxx:


synopse:
I merged the request, but this implementation is not safe at all.
A ID value collision may occur very easily, if the ID is increased outside the mORMot core in between maxID computation and document insertion.

zedxxx:
And how do it safe?

synopse:
I guess the only solution is to use a centralized service to compute 
increasing IDs for ALL process accessing to the MongoDB engine.
OR retry until it worked...

But IMHO centralizing all MongoDB access at mORMot level is the best 
idea.

---
Arnaud Bouchez
http://synopse.info

Le 2015-11-02 15:09, zedxxx a écrit :

PR #9, opened 2015-12-17
Added bswap16 function

laurent-favreau:


synopse:
Closed pull request since it is a built-in function.

PR #13, opened 2016-03-18
add lazarus support for sample 10

mariuszekpl:


mariuszekpl:
WinSvc unit is unused and not exists in Lazarus

PR #16, opened 2016-07-14
Path separator must be "/"

5050:
From ZIP file specification 2014 https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT, chapter 4.4.17:
- it is forbidden to have a drive letter in path
- path separator MUST be "/" and not "\"

The patches reverses the change from '/' to '\' for others OS than windows to a change from '\' to '/' for windows and replace drive letter by a letter path


synopse:
AFAIR there may be some incompatibility issues if you follow this pattern.

PR #18, opened 2016-07-14
Path separator must be "/", drive letter is forbidden 

5050:
From ZIP file specification 2014 https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT, chapter 4.4.17:

    path separator MUST be "/" (and not "\")
    it is forbidden to have a drive letter in path

The patches reverses the change from '/' to '\' for others OS than windows to a change from '\' to '/' for windows and replace drive letter by a letter path

### DIFF
--- a/SynZip.pas
+++ b/SynZip.pas
@@ -1019,13 +1019,14 @@ begin
   try
     if ZipName='' then
       if RemovePath then
-        ZipName := ExtractFileName(aFileName) else
+        ZipName := ExtractFileName(aFileName)
+      else
         ZipName := aFileName;
     {$ifdef MSWINDOWS}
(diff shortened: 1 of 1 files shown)


synopse:
AFAIR the APPNOTE is not followed by some implementations, especially under Windows.
If you put "/" as required, some readers (e.g. the embedded .zip folder in older versions of Windows if I remember well), may not work as expected.
And letter drives would never appear, since all those folder names should be relative.
Have you any difficulty with the current pattern, which is not following the APPNOTE, but was reported to work in practice.
It has to be cross-checked - I do not want to merge this request without the assurance it won't break anything.

synopse:
No reaction since my answer of July 18. So I guess nothing to change.

PR #19, opened 2016-08-29
More proper compilation of SynSQlite3Static for FPC (use "public name…

maciej-izak:
…" instead of deprecated "alias:"). No need for declaration in interface section.


synopse:
I did a manual merge.

PR #21, opened 2016-10-21
Add Support for ARC ARCTO PIE and CHORD

ProHolz:
I have added Support for missing functions.  
If there are any problems with these please contact me
[TestMetaCanvas.pdf](https://github.com/synopse/mORMot/files/545125/TestMetaCanvas.pdf)


ProHolz:
Removed unit SynPdfArc

PR #22, opened 2016-10-29
varFmtBCD support in TTextWriter.AddVariant

ppm0:
please add...


synopse:
Problem with this kind of implementation is that it will add a dependency to Data.DB and Data.FMTBcd units in SynCommons, which is not what we expect!

synopse:
I've implement it properly with a variant-dedicated serialization mechanism.
See https://github.com/synopse/mORMot/commit/8987888f72e0b988f176952d90cb258614b3c603

PR #37, opened 2017-04-18
Converting ModTime created a wrong datetime error

martin-doyle:
Update SynVirtualDataSet.pas
Update GetFieldData ftDateTime
Update mORMot
Update EngineDelete


martin-doyle:
I agree this is not a solution but a workaround. 
The problem happens in the rare event that a delete is the first transaction after starting the program and the deleted record has the highest version number. I will write a test which shows that.

martin-doyle:
The test is uploaded to the new project https://github.com/martin-doyle/SynopseTestReplication. It failes if the last added record is deleted. I reloaded the TSQLRestServerDB before the delete in order to simulate a new program start.

martin-doyle:
@synopse Did you have a chance checking the test at  https://github.com/martin-doyle/SynopseTestReplication?

PR #43, opened 2017-06-21
Internalstate support for retrievelist

ASiwon:
Please add filling InternalState property for RetrieveList* methods.


ASiwon:
Thank you very much for information. I want to fill InternalState property for all TSQLRecord items in the list and in this case assignment should be done inside of the loop. I made copy/paste error using Item variable in nongeneric version which is corrected now.

PR #44, opened 2017-06-25
Revert commit: try to circumvent weird XE4/XE5 compiler issues.

ASiwon:
Revert commit: try to circumvent weird XE4/XE5 compiler issues. Problem was due to wrong Synopse.inc content in local copy. Now, when the reason of problem is known, change could be reverted.


synopse:
No need to revert it, in fact. It won't hurt to publish it in the implementation section.

PR #45, opened 2017-06-27
fix unhandled helpers without parameters like NewGUID. Now is possibl…

maciej-izak:
…e for example to execute {{NewGUID}}


synopse:
your commit did break the regression tests - I will fix it

PR #46, opened 2017-08-16
Windows 8.1 and newer. Use WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY

cybexr:


synopse:
You introduced a dependency into SynCommons, which was to be avoided in this unit.
I will include the patch by using another way for finding out the OS revision. ;)

synopse:
Included as https://github.com/synopse/mORMot/commit/a2a23b0ed8c111d7fed3dc0618844fb294bf8a0e

PR #47, opened 2017-09-01
-added: Sample 31 for Lazarus (converted from the Delphi one) ChatSer…

afriess:
one sample more working with Lazarus.


synopse:
Outdated.

Please reopen if needed.

PR #54, opened 2017-11-07
File version info is loaded in OS independed way

ssoftpro:
Starting from version 3.0 FPC stores file version info on different platforms. ExeVersion.Version has been extended to support this


SynCommons.pas
```diff
@@ -36863,11 +36873,59 @@ constructor TFileVersion.Create(const aFileName: TFileName;
     end;
   end;
   {$endif}
-  SetVersion(aMajor,aMinor,aRelease,aBuild);
-  if fBuildDateTime=0 then  // get build date from file age
+  {$if defined(FPC) and defined(VER3_0) not defined(MSWINDOWS)} // Only works starting from FPC 3.0
```
pavelmash: "and" is missing here. Should be "not defined(MSWINDOWS)"

SynCommons.pas
```diff
@@ -36815,6 +36820,11 @@ constructor TFileVersion.Create(const aFileName: TFileName;
         SetRawUTF8(Result,StrValPt,sz)
     end;
 {$endif}
+{$if defined(FPC) and defined(VER3_0) and not defined(MSWINDOWS)}
+  VI: TVersionInfo;
+  LanguageInfo: String;
+  TI, I: Integer;
+{$endif}
```
pavelmash: Should be {$ifend} (paired  to {$if})

ssoftpro:
I've added ISFPC30 definition to Synopse.inc to avoid extended conditionals in SynCommons

PR #58, opened 2017-11-30
Fix for Android compilation problem (only TSQLHttpClientWinSock is al…

maciej-izak:
…lowed)


synopse:
Please check https://github.com/synopse/mORMot/commit/d01d9d6c96baff611c86840ef3e4c4509d124bbf

PR #66, opened 2017-12-13
Fix: fOnTerminate event was set too late

ssoftpro:
An wasn't propagated properly to working threads


pavelmash:
@synopse PLease, review this pull request. I can confirm - current implementation don't assign a TSynThreadPoolSubThread.OnTerminate value from Owner. We need to shut down JSEngine/OLEDB connection in the same thread it created and use a OnTerminate event for this. For HTTP API server all works correctly, for sockets server this patch is required

PR #67, opened 2017-12-21
ThreadID is 8 bytes long under Linux64

ssoftpro:


pavelmash:
Nice catch! But it is better to use Writer.AddQ -> Writer.Add. Writer.AddQ will broke windows. Please, fix this and I will merge

ssoftpro:
Done

PR #69, opened 2017-12-24
ResolveIPToName implementation is aligned in SynWinSock and SynFPCSock

ssoftpro:


pavelmash:
@synopse - please, review this change - i't allow to write the same code for fpc<->delphi

PR #74, opened 2018-01-06
Add subscript support to TDocVariant

zhouzuoji:
Add subscript support to TDocVariant
Add OLE array support for TTextWriter.AddVariant

here is a test:
procedure func1;
var
  info: Variant;
begin
  info := _JsonFast('{"students":[{"age":10,"ismale":false,"name":"Susan","address":{"state":"Washington","city":"Seattle"}},{"age":11, "name":"Alex","ismale":true}]}}');
  Writeln(info.students[1]);  
  Writeln(info.students[0].address['state']);
end;

### DIFF
--- a/SynCommons.pas
+++ b/SynCommons.pas
@@ -788,6 +788,7 @@ uses
 {$endif}
 {$ifndef NOVARIANTS}
   Variants,
+  VarUtils,
 {$endif}
   SynLZ, // needed for TSynMapFile .mab format
   SysUtils;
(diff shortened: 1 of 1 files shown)


synopse:
Nice!
Is it working from Delphi 6 to Delphi 10.2 too?

zhouzuoji:
Sorry, I test it only on Delphi 2010. 
I use Delphi 2007 occasionally， never use other delphi versions and FPC.
Just now,  I found some compiling error on Delphi 2007, because it has no Variant-Type VarUString, it's easy to solute.

TSynInvokeableVariantType.DispInvoke is copied Variants.TInvokeableVariantType.DispInvoke, it may differ in different versions. I will try to fix it.

zhouzuoji:
It seems that your TDocVariant do not support FPC, not because my modification

info := _JsonFast('{"students":[{"age":10,"ismale":false,"name":"Susan","address":{"state":"Washington","city":"Seattle"}},{"age":11, "name":"Alex","ismale":true}]}}');
Writeln(info); // this works
Writeln(info.students); // this cause crash

synopse:
It works with newer version of FPC, with some minor incompatibility problems sometimes.
The easiest with FPC is to use the TDocVariantData - by calling _Safe() - directly.

zhouzuoji:
I tested this in Lazarus Release 1.8 (with FPC 3.0.4)  Win64, it still raise an access violation exception and crash. maybe FPC does not support Variant method calling

PR #75, opened 2018-01-08
subscript-access support for custom variant (TDocVariantData)

zhouzuoji:
Add subscript-access support for custom variant (TDocVariantData).
Add rtl built-in variant array support for TTextWriter.AddVariant.
It works on lazarus 1.8 (win32 and win64), Delphi7, Delphi 2007, Delphi2010 now,
Other platforms are not testd, but  I think it works on higher Delphi version too.
It's hard to do test on such many platforms, is there an automatic-testing method?

[here is my test code](https://github.com/zhouzuoji/mORMotTest/tree/master/DocVariant)


zhouzuoji:
I tested on Ubuntu/lazarus 1.6.4 and Delphi XE6, it works too.
Delphi XE 10.2 can't compile, because Variants.TCustomVariantType's virtual method DispInvoke has such a prototype:
  procedure DispInvoke(Dest: PVarData; [Ref] const Source: TVarData; CallDesc: PCallDesc; Params: Pointer); override;
a [Ref]  is added to decorate the const parameter "Source"

just modify SynCommons.TSynInvokeableVariantType.DispInvoke's prototype and it works on Delphi XE 10.2 too.

I do not known [Ref] is added after which version, otherwise it's easy to make it compatible with Delphi XE 7 and higher versions. I need your help.

PR #76, opened 2018-01-21
Fixed a bug in TSQLPropInfoRTTIDynArray.SetValue

transmogrifix:
Fixed a bug in TSQLPropInfoRTTIDynArray.SetValue which didn't try to decode base64 input if its fObjArray was assigned.

### DIFF
--- a/SQLite3/mORMot.pas
+++ b/SQLite3/mORMot.pas
@@ -23194,9 +23194,11 @@ begin
   if Value=nil then
     da.Clear else
     try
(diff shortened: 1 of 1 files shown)


transmogrifix:
I have explained this at https://synopse.info/forum/viewtopic.php?id=4339
I have also provided a minimal example which reproduces the problem.

transmogrifix:
The root cause has been corrected (https://synopse.info/fossil/info/e7d106ff7a), which makes this pull request redundant.
Thanks ab.

PR #77, opened 2018-01-25
WIP: THttpServer fixes

ssoftpro:


ssoftpro:
Unfortunately no, exit<>break because we are in the second internal loop. Exit forces control to leave the subroutine and go back to the calling routine, while break just stops execution of the second loop and continues execution of the first loop.
By the way, I've closed this pool request in order to create another one with more changes in it

PR #78, opened 2018-01-25
An ability to disable HTTP/1.1 KeepAlive support added

ssoftpro:


SynCrtSock.pas
```diff
@@ -1843,6 +1845,11 @@   THttpServer = class(THttpServerGeneric)
     // a THttpServerResp thread is created for handling this THttpServerSocket
     property Sock: TCrtSocket read fSock;
   published
+    /// flag to disable HTTP/1.1 keep alive features and fall back to
+    // HTTP/1.0 compatibility
+    // Only threads from thread pool will be used to process requests
+    // when this flag set to true
+    property DisableKeepAliveSupport: boolean read fDisableKeepAliveSupport;
```
synopse: add "write" attribute too?

SynCrtSock.pas
```diff
@@ -1843,6 +1845,11 @@   THttpServer = class(THttpServerGeneric)
     // a THttpServerResp thread is created for handling this THttpServerSocket
     property Sock: TCrtSocket read fSock;
   published
+    /// flag to disable HTTP/1.1 keep alive features and fall back to
+    // HTTP/1.0 compatibility
+    // Only threads from thread pool will be used to process requests
```
synopse: missing traiiling "-" character in comment

SynCrtSock.pas
ssoftpro: Let's don't forget that this is a thread so execution starts immediately. It could be too late to set this flag after execution started and first requests came
On the other side, IMHO it is non-practical to switch such behavior on the fly.

ssoftpro:
The primary point of change in the loop is to change this part:
    if fThreadPoolPush(pointer(ClientSock)) then
       exit; // the thread pool acquired the client sock
'exit' here results in stop execution of the Execute method in whole, so server stops it's execution at all
I would appreciate if you have a better solution for this.

SynCrtSock.pas
ssoftpro: Change committed. Did I understand you correctly?

synopse:
now I understood the problem of the threadpool contention loop !
this was indeed a nasty bug

synopse:
I've made some changes to how the contention is handled - especially changed ThreadPoolContentionCount into ThreadPoolContentionTime and enhanced the doc

synopse:
the commit also broke THttpServerSocket.GetRequest with no associated server
(it made a GPF and failed TTestProtocols.RTSPOverHTTP) 
-> please ensure in the future that you run TestSQL3 regression tests before any pull request ;)

PR #79, opened 2018-01-31
TObjProxyVariant - a custom variant type used to have direct access t…

maciej-izak:
…o object published properties. Very usable for cases where conversion from object to variant is not sufficient. TObjProxyVariant provides direct access to object properties from Variant instead of conversion to Variant.


maciej-izak:
Thanks!

PR #80, opened 2018-02-06
fix incomlete implementation of TPropInfo.isStored for FPC

pavelmash:
FPC implementation of mORMot.TPropInfo.IsStored is incompleate for FPC compiler (handle only consts), so for case
```
TBaby = class
  private
    fName: RawUTF8;
  function canStoreName: boolean;
  published
    property Name: RawUTF8 read fName write fName stored canStoreName;
end;
```
Name property will never be stored


synopse:
Has been fixed as part of https://synopse.info/fossil/info/e58267d974

PR #81, opened 2018-02-10
Introduce THttpServerGeneric.OnBeforeRequestProcessed & OnAfterReques…

pavelmash:
Introduce THttpServerGeneric.OnBeforeRequestProcessed & OnAfterRequesProcessed events. Implement it for HTTP.SYS based server. See [this forum post for details](https://synopse.info/forum/viewtopic.php?pid=20532#p20532)

Actual changes is very small - I just move a peace of code for sending response to SendResponse without any modification. 

Need help to implement the same for socket based server

### DIFF
--- a/SynCrtSock.pas
+++ b/SynCrtSock.pas
@@ -1134,6 +1134,8 @@ type
     /// optional event handler for the virtual Request method
     fOnRequest: TOnHttpServerRequest;
     fOnBeforeBody: TOnHttpServerBeforeBody;
+    fOnBeforeRequestProcessed: TOnHttpServerRequest;
+    fOnAfterRequestProcessed: TOnHttpServerRequest;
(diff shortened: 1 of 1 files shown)


synopse:
Sounded better to me to do this at TSQLHttpServer.Request + TSQLRestClientURI.URI, using a TSQLRestURIParams callback.
So that the very same callback will be usable on both client and server sides.

pavelmash:
Unfortunately this doesn't meet my condition. My code work on the THttpServerGeneric level. 5 years ago TSQLHttpServer was not so mature as today and I start my code from plain http server. There is no way back for me..

synopse:
Please check https://github.com/synopse/mORMot/commit/dc316fd59565d97823a15c796e5e3ef7c77e54ec implementation

pavelmash:
Thank you! Works as expected for me

PR #82, opened 2018-02-15
SynZip: use a SynZLibSSE by default for FPC/Win64

pavelmash:
because of (possible) bug in fpc paszlib - see [this forum thread](https://synopse.info/forum/viewtopic.php?id=4400)

### DIFF
--- a/SynZip.pas
+++ b/SynZip.pas
@@ -177,7 +177,7 @@ unit SynZip;
       {.$define USEZLIBSSE} // SynZLibSSE static .o files for FPC + Win32 fails
     {$endif}
     {$ifdef Win64}
-      {$define USEZLIBSSE} // SynZLibSSE static .o files for FPC + Win64
+      {.$define USEZLIBSSE} // SynZLibSSE static .o files for FPC + Win64
     {$endif}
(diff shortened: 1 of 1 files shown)


pavelmash:
Actual [problem is inside SynZip.GZRead. Fixed.

synopse:
My guess is that inflate() has problem when bits = -MAX_WBITS in the FPC pascal zlib.

pavelmash:
I understand this is dirty hack. You are right about gzlen should be - 18. But I analyze how `fpcsrc\packages\paszlib\examples\minigzip.pas` sample work.  It use `bits = -MAX_WBITS` during inflateInit2_ and pass a length of last chunk as in my fix (length of all data in buffer including 8 bytes with crc & len) .

synopse:
Please check https://github.com/synopse/mORMot/commit/3b7015f9bdc60a2e7a16b3257436f5ff58e03490
I hope this branch is useless now.

pavelmash:
Closed because of [SynZip refactoring](3b7015f9bdc60a2e7a16b3257436f5ff58e03490)

PR #83, opened 2018-02-23
Global proxy is initialized from HTTP_PROXY environment variable

ssoftpro:


pavelmash:
I commit this to trunk instead of fb_SyNode_fpc311 brunch

PR #84, opened 2018-02-23
Fix: GET after POST reqults in POST sent

ssoftpro:
When the same TCurlHttp instance is used to send multiple requests and POST request is followed by GET request, POST was sent to server


pavelmash:
I merge this to trunk instead of fb_SyNode_fpc311 brunch

PR #87, opened 2018-02-27
Virtual methods for loading templates in MVC Web App

achechulin:
Customizable loading TMVCApplication template and static files (from application resources, from zip archive and etc).
Example:
https://synopse.info/forum/viewtopic.php?id=4066


achechulin:
I tried to maintain maximum backward compatibility.

Differenses:
1. FindTemplates does not return files marked as `faSysFile` and `faHidden` (same as FindFiles).
2. In TMVCViewsMustache.GetRenderer the text of the exception message contains ShortFileName instead of FileName. It has done so because the FileName is long and hard to watch (in my case). It's ok to change back to FileName.

PR #88, opened 2018-02-28
fix exception logging for Win+FPC

pavelmash:


pavelmash:
On FPC version <  3.1.1-r38162 ( I test on 3.0.2 & 3.0.4) function ExceptionInheritsFrom raise a SIGSEGV error in case exception class is inherited from something - in my case raised EMetabaseException = class(ESynException). After we calculate parent using:
```
E := PPointer(PtrInt(E)+vmtParent)^;
```
This line fails:
```
IdemPropName(PShortString(PPointer(PtrInt(E)+vmtClassName)^)^,Name)
``
May be you can fix it on RTTI level?

synopse:
Please check https://github.com/synopse/mORMot/commit/c38bee17c08799b6d33fbd0a9dbeb6e0bfed0b8b and https://synopse.info/forum/viewtopic.php?pid=26675#p26675

pavelmash:
Closed in favor of [c38bee1]

PR #90, opened 2018-03-01
fix surrogates in RawUnicodeToUtf8 for FPC

pavelmash:
Test case is string with broken surrogate pair 'ab\ud800cd'
FPC compiler (x64) evaluated condition
```
cardinal(Source^)-UTF16_LOSURROGATE_MIN>UTF16_LOSURROGATE_MAX-UTF16_LOSURROGATE_MIN
```
to false, Delphi - to true.
Condition replaced by more compiler (and human) friendly


pavelmash:
Will be good to add a test case to SynSelfTests.pas for this....

PR #92, opened 2018-03-02
On non-Windows systems GetSystemCodePage function returns code page

ssoftpro:
This function was introduced in FPC 2.7.1

### DIFF
--- a/SynCommons.pas
+++ b/SynCommons.pas
@@ -832,7 +832,7 @@ const
 
 {$ifndef MSWINDOWS}
   /// estimate the system code page is WinAnsi
-  GetACP = CODEPAGE_US;
+  GetACP: TSystemCodePage = CODEPAGE_US;
   /// internal Code Page for UTF-8 Unicode encoding
   CP_UTF8 = 65001;
 {$endif}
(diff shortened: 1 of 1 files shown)


pavelmash:
@synopse Please, apply this patch. Without it CurrentAnsiConvert works incorrectly for non-latin characters in most Linux (because default code page in most case is UTF-8 )

synopse:
Please add a conditional for $ifdef ISFPC271 then I will merge the patch ;)

pavelmash:
ISFPC271 is already defined by [Synopse.inc:336](https://github.com/synopse/mORMot/blob/34c893a17f1804b090c2ac350663cd0dfbe2455a/Synopse.inc#L336)

pavelmash:
@synopse - maybe I misunderstood you about "Please add a conditional for $ifdef ISFPC271"? Please, give me a feedback. This patch is critical for me - we have a cyrillic characters in most of our files

synopse:
Check https://synopse.info/fossil/info/67711926bf

pavelmash:
Verified, work as expected. Thanks!

PR #96, opened 2018-03-10
fix TOracleDate zero comparision; remove FPC hints

pavelmash:
Please, review my replacements of  PInteger(PtrInt(@self)+3)^ -> PInteger(PByte(@self)+3)^. Seems PtrInt(@self)+3 is wrong..


synopse:
replacing PtrInt by PByte doesn't compile with Delphi in such expression
this is why we didn't make the change already

PR #97, opened 2018-03-15
Fix SynOleDB for FPC & x64 target

pavelmash:
This patch fix SynOleDB for FPC & x64 target:
 - FPC don't understand {$A+} {$A-} instruction, replaced by {$A8} {$A1} 
- ULONG/TUIntDynArray replaced by PtrUInt/TPtrUIntDynArray where expected
- `packed` keyword removed from records for proper alignment
- minor refactoring of Int64/UTF8 array binding

Verified with XE2 Win32 & FPC311 Win64 (hope also work with Delphi x64)

### DIFF
--- a/SynOleDB.pas
+++ b/SynOleDB.pas
@@ -295,6 +295,10 @@ type
   DBPARAMFLAGS = DWORD;
   DBTYPE = Word;
   DBRESULTFLAG = UINT;
+
+  DBLENGTH = PtrUInt;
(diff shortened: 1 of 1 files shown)


synopse:
Delphi 5 doesn't understand the {$A1} directive, so I've changed it into {$A-} and I hope it works as expected with FPC (seems OK from my own tests)

PR #100, opened 2018-03-26
initialize a new temporary buffer with default size of 4KB

maciej-izak:


synopse:
but why? just define `tmp: array[0..4095] of AnsiChar;` in your local stack

PR #101, opened 2018-03-27
Fixed support of mssql connection using odbc

ssoftpro:
Verified on Linux connecting to Linux version of MSSQL


synopse:
BindArray for ODBC is just a **great** feature!
Thanks!

synopse:
BTW why is it allowed only for MSSQL? Perhaps other providers do support it.

Also note that you used StrLCopy which is not available e.g. on non Unicode Delphi, so I will fix it.

synopse:
Please check https://github.com/synopse/mORMot/commit/1cfd6bb31e8e3cc9c153914d985188606fd2dc73

PR #102, opened 2018-03-28
Fixed: TSynAnsiUTF8.UnicodeBufferToAnsi resulted in truncated string

ssoftpro:
Despite double sized buffer allocated for conversion, only SourceChars where used resulting in truncated string when non Ascii characters present in Source.
To be safe I propose to pass an actual size of buffer allocated into UnicodeBufferToAnsi function
Only the change to TSynAnsiUTF8.UnicodeBufferToAnsi is mandatory withal

### DIFF
--- a/SynCommons.pas
+++ b/SynCommons.pas
@@ -1261,9 +1261,9 @@ type
     // - Dest^ buffer must be reserved with at least SourceChars*3 bytes
     // - this default implementation will rely on the Operating System for
     // all non ASCII-7 chars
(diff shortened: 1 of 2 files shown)


pavelmash:
@ssoftpro This is a huge breaking changes. To fix  RawUnicodeToString for case system code page is UTF8 better mark a 
```
TSynAnsiConvert.UnicodeBufferToAnsi(Source: PWideChar; SourceChars: Cardinal)
```
as virtual and implement it in
``` 
function TSynAnsiUTF8.UnicodeBufferToAnsi(Source: PWideChar; SourceChars: Cardinal): RawByteString;
```

PR #103, opened 2018-04-04
TCurlHTTP: POST request with empty body can return data

pavelmash:
- simplify code
- fixed: non-GET request with empty body can return data (CURLOPT_NOBODY should be set only for HEAD requests)
- fixed: remove 1-2s delay while sending non-GET request with empty body 
all mormot test are passed under Linux

### DIFF
--- a/SynCrtSock.pas
+++ b/SynCrtSock.pas
@@ -11026,20 +11026,6 @@ begin
   end;
 end;
 
-function CurlReadData(buffer: PAnsiChar; size,nitems: integer;
-  opaque: pointer): integer; cdecl;
-var instance: TCurlHTTP absolute opaque;
-    dataLen: integer;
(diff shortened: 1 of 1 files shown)


synopse:
why did you remove https://curl.haxx.se/libcurl/c/CURLOPT_READDATA.html ?
are you sure your modification works with a GET including a body?
IIRC this function was introduced for such a case

pavelmash:
With CURLOPT_READDATA  POST request without body ( in fact any  non-GET
request without body) wait for 2-3 sec to send data.
Yes, I'm sure GET with body work correctly - this error was found by my
autotests (and fixed by ssoftpro) so I verify this case..

2018-04-05 9:40 GMT+03:00 ab <notifications@github.com>:

PR #104, opened 2018-04-05
Support for XE4 for SynTaskDialog4Lazarus

yonojoy:
ThirdPartyDemo SynTaskDialog4Lazarus did not compile under XE4/FireMonkey. This patch allows the usage under XE4/FMX. Scaling however will use defaults because of missing functionality in XE4-FMX


synopse (approved):
Why not?

yonojoy:
> Why not?

Why scaling will not work?

Scaling for non-Windows platforms depends on `TDeviceDisplayMetrics` which is available since XE8 (see `ScalingByScreenDPI_N`). However, for Windows scaling should work.

PR #106, opened 2018-04-18
TPollAsynchSockets disconnect on error and store WSAGetLastError code

f-vicente:
According to the documentation, after a Send call only WSAEWOULDBLOCK is recoverable trying later, other errors should close the connection.

Also added lastWSAError field to the TPollSocketsSlot, to store the error code and be able to process later (i.e. on TAsynchConnection.BeforeDestroy).

Tried the code disconnecting clients from a server and successfully got 10054 (connection reset) or 0 (closed gracefully) if the client disconnects abruptly or not.


synopse:
sounds just fine!
thanks for sharing

f-vicente:
Thnanks! 

But I think that I've uploaded a wrong change to the ConnectionRemove function. 

I was having problems removing connections from other thread, so I implemented a lock on the ConnectionRemove function, but since the TPollSocketsSlot:Lock works increasing/decreasing a counter, this change will break the ConnectionRemove if called inside TAsynchConnection.OnRead method as the comment says.

Could you revert the changes on SynBidirSock.pas? the ones on SynCrtSock seems to work fine.

PR #107, opened 2018-04-21
TSynMemoryStreamMapped constructor overloaded to prevent conflict wit…

pavelmash:
In case
```
constructor TSynMemoryStreamMapped.Create(const aFileName: TFileName;
  aCustomSize: cardinal; aCustomOffset: Int64);
```
called with one argument (aFileName) then actual stream is created using
```
constructor TSynMemoryStream.Create(const aText: RawByteString);
```
and content of a stream is a file name, instead of file content

### DIFF
--- a/SynCommons.pas
+++ b/SynCommons.pas
@@ -10355,6 +10355,9 @@ type
   end;
 
   /// a TStream created from a file content, using fast memory mapping
+
+  { TSynMemoryStreamMapped }
+
   TSynMemoryStreamMapped = class(TSynMemoryStream)
   protected
     fMap: TMemoryMap;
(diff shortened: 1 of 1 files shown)


synopse:
Your patch uses `aFileName: RawByteString` which seems weird/incorrect, and I'm not sure there won't be any conflict when calling it: I guess the compiler may complain about ambiguous call...

pavelmash:
Here is the problem in my code (FPC) I try to solve with this patch ( fnUTF is come from JSON as RawUTF8)
```
var
  fnUTF: RawUTF8;
begin
  fnUTF := 'compil.bat';
  Stream := TSynMemoryStreamMapped.Create(fnUTF); // stream contains string 'compil.bat'
```
I know this is my mistake and I replace code by valid one:
```
  Stream := TSynMemoryStreamMapped.Create(UTF8ToString(fnUTF)); // stream contains content of file
```
But IMHO behavior is unexpected

pavelmash:
Closed. developer should be careful while using compiler typecasted parameter in TSynMemoryStreamMapped.Create

PR #108, opened 2018-05-05
Fixed a bug in TSQLPropInfoList.InternalAddParentsFirst

transmogrifix:
Fixed a bug in TSQLPropInfoList.InternalAddParentsFirst which caused all properties to be skipped after a property with a getter is encountered and pilIgnoreIfGetter is in fOptions.


synopse:
Thanks for the report!

PR #110, opened 2018-05-08
GSSAPI support added

ssoftpro:
Verified on Linux


synopse:
Nice.
What about the integration with the mORMot authentication scheme?

ssoftpro:
I've added support on a server for browser clients due to lack of required functions in SynGSSAPI
Successfully checked with Project04Server example on Linux while a client was a Chrome browser on a Windows-based computer.

ssoftpro:
A few additional libraries and comprehensive configuration required to successfully run this. So I do not recommend to enable this feature by default

synopse:
To prevent linking issues at compilation time, and enable it by default, you may consider dynamic loading the library.

ssoftpro:
Done

PR #115, opened 2018-05-30
Boolean Handling

MartinEckes:
The are two Issues with the boolean handling. Please correct the source formating, this is my first time.


MartinEckes:
Here is my next try.

PR #116, opened 2018-06-02
Few changes for Unix and Android.

LongDirtyAnimAlf:
See comments of commits. Thanks !

### DIFF
--- a/SQLite3/mORMotUI.pas
+++ b/SQLite3/mORMotUI.pas
@@ -128,7 +128,12 @@ interface
 {$I Synopse.inc}
 
 uses
+  {$ifdef MSWindows}
   Windows,
+  {$endif}
+  {$ifdef FPC}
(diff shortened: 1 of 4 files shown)


synopse:
Thanks a lot for sharing!
I've merged it but I wonder why is the sqlite3.o so big? It is more than 5MB whereas the arm-linux version you supplied some months ago (IIRC) is less than 1MB.

LongDirtyAnimAlf:
I saw it ... and I do not know why !
I just the "ndk-build.cmd" from the Android toolchain, together with a tuned "Android.mk" build-file for sqlite3.
Settings regarding debugging or optimization level do not have that much influence on the filesize unfortunately.

LongDirtyAnimAlf:
Sidenote: the resulting executable (so, generated with FPC and LAMW) only grows with less than 1MB !!

PR #118, opened 2018-06-10
Some fixes for Darwin

LongDirtyAnimAlf:
Latest mORMot did not compile anymore on Darwin. These proposed fixes do solve.

### DIFF
--- a/SynZip.pas
+++ b/SynZip.pas
@@ -373,12 +373,13 @@ type
         libz='zlib-64.dll'; // as available in \fpc-win64 sub-folder
         {$endif}
       {$endif MSWINDOWS}
-      {$ifdef LINUX} // dynamically linked with new 64-bit TZStream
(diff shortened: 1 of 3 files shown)


SynZip.pas
```diff
@@ -373,12 +373,13 @@   ESynZipException = class(Exception);
         libz='zlib-64.dll'; // as available in \fpc-win64 sub-folder
         {$endif}
       {$endif MSWINDOWS}
-      {$ifdef LINUX} // dynamically linked with new 64-bit TZStream
+      {$ifdef UNIX} // dynamically linked with new 64-bit TZStream
```
synopse: will it continue to work on Linux / BSD too ?

LongDirtyAnimAlf: Yes !
When Linux is defined, Unix is also defined.
When Darwin (BSD) is defined, also Unix is defined.

PR #122, opened 2018-07-17
Small improvements to SyNode

ssoftpro:
1) jsid redefinition for easier use in some cases
2) small nonfunctional change in SyNode remote debugger


SyNode/SyNodeRemoteDebugger.pas
```diff
@@ -537,7 +539,7 @@ procedure TSMRemoteDebuggerCommunicationThread.HandleMessage(const request: Vari
 procedure TSMRemoteDebuggerCommunicationThread.Send(const packet: RawUTF8);
 begin
   sockWrite(packet);
-  SynSMLog.Add.Log(sllCustom4, packet);
+  SynSMLog.Add.Log(sllCustom3, 'JSDBG packet sent: %', packet);
```
pavelmash: Please, don`t change the log level of JSDBG protocol. The JSDBG log is huge and apps can disable a sllCustom4 to decrease logs size. For example in UnityBase app we enable sllCustom4 only in case FF_RDP_DEBUG enviroment variable is defined
```
      // enable FireFox debugger protocol logging only if FF_RDP_DEBUG
      // environment variable is defined
      // this prevent a HUGE log files
      if (GetEnvironmentVariable('FF_RDP_DEBUG')='') then
        Level := Level - [sllCustom4];
```

ssoftpro:
Log level changes removed

PR #123, opened 2018-07-19
TSynTableStatement add OrderByDescField property, contains desc-ordered field-index.

cybexr:
Fix order by SQL with multifields and desc BUG, eg: order by fn1 desc, fn2, fn3 desc, fn4 asc limit 10
test with  D7&fpc on win64.


synopse:
Has been properly implemented in mORMot 2.

PR #124, opened 2018-08-15
Linux compatibility fix

ssoftpro:
This fixes directory support in SynZip - forces correct direction of slashes (forward <-> backward), depending on OS a program is running on

### DIFF
--- a/SynZip.pas
+++ b/SynZip.pas
@@ -1473,8 +1473,12 @@ var Parent: TFileName;
 begin
   result := false;
   if Path='' then exit;
+  {$IFDEF FPC}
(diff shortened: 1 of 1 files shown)


pavelmash:
@synopse - please, apply this patch - we need to work with folders in zip archives on linux

synopse:
IncludeTrailingPathDelimiter() is enough on all targets, I guess

ssoftpro:
Unfortunately no. I've run our tests on Linux to unzip archive created on Windows and got files named like 'dir\subdir\name.ext' right in the root folder

PR #128, opened 2018-08-31
Fix properties OnBeforeRequest and OnAfterRequest for THttpServerGeneric

maciej-izak:


synopse:
nice finding!
some others are needed... I think!

PR #141, opened 2018-10-14
add ECC.lpi

pavelmash:
I'm push this to fossil trunk but unable to push directly to github master - some pre-commit hook  require at last one reviewer


pavelmash:
UPD. No - I didn't push to fossil.  Please, move it there after merging

PR #144, opened 2018-10-20
Fix compilation with FPC using build-fpc-linux64.sh

michaliskambi:
Three straightforward fixes (two to build-fpc-linux64.sh script, one to SynSelfTests.pas) :) See the commits for details.

Compilation like

```
./SQLite3/build-fpc-linux64.sh
```

works OK now with FPC 3.3.1-r39998.


synopse:
I suspect some latest commits by mpv did make the trick

PR #146, opened 2018-10-29
support for non-english resourcestrings, no need to restart an app after change of language, support for build under FMX

paustr:
1. Extract non English resourcestrings into .messages files (used rawUTF8 type istead of Ansi)
2. Change language on the fly without restarting an application, many times
- added {$undef USEFORMCREATEHOOK}, to make SetCurrentLanguage procedurę available
- modified procedurę SetCurrentLanguage  
3. some FMXAPP directives in order to be able to build under fmxApplication
 
Sample application
https://github.com/paustr/multiLangApp

### DIFF
--- a/SQLite3/mORMoti18n.pas
+++ b/SQLite3/mORMoti18n.pas
@@ -2261,7 +2261,7 @@ begin
   if (PtrInt(lpszType)<>PtrInt(RT_STRING)) then exit;
   i := (PtrInt(lpszName)-1)shl 4;
   for i := i to i+15 do begin // resourcestrings are stored by groups of 16
-    SetString(s,buf,LoadStringW(hInstance,i,buf,sizeof(buf)));
+    s := RawUnicodeToUtf8(buf,LoadStringW(hInstance,i,buf,sizeof(buf)));
     if s='' then exit; // we reach the end
     AddOnceDynArray(s);
   end;
--- a/SQLite3/mORMoti18n.pas
+++ b/SQLite3/mORMoti18n.pas
@@ -809,6 +809,11 @@ begin
     result := TLanguages(ndx);
 end;
 
+function Hash32(const Text: RawUTF8): cardinal;
+begin
+  result := SynCommons.Hash32(pointer(Text),length(Text));
+end;
+
 
 const
   // default character set for a specific language (for GUI i18n)
@@ -1931,10 +1936,8 @@ procedure TLanguageFile.Translate(var English: string);
 // case-sensitive (same as standard gettext)
 var result: string;
 begin
-  //resourcestrings may be not in English
   result := FindMessage(Hash32(English)) ;
-    // resourcestring are expected to be in English, that is WinAnsi encoded
- //   {$ifdef UNICODE}StringToWinAnsi{$endif}(English)));
+
   if result<>'' then
     English := result;
 end;
@@ -2229,7 +2232,7 @@ var
 
 function Hash32Str(crc: cardinal; buf: PAnsiChar; len: cardinal): cardinal;
 begin
-  result := Hash32(buf,len);
+  result := SynCommons.Hash32(buf,len);
 end;
 
 function AddOnceDynArray(const S: RawUTF8): integer;
--- a/SQLite3/mORMoti18n.pas
+++ b/SQLite3/mORMoti18n.pas
@@ -211,14 +211,21 @@ interface
   {$undef USEFORMCREATEHOOK}
 {$endif}
 
+{.$define FMXAPP}
+//must be set globally to use this unit in FireMonkey application
+
 uses
   Windows, SysUtils, Classes,
   {$ifdef USEFORMCREATEHOOK}
   {$ifndef LVCL}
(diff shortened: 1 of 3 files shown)


SQLite3/mORMoti18n.pas
```diff
@@ -1907,9 +1931,10 @@ procedure TLanguageFile.Translate(var English: string);
 // case-sensitive (same as standard gettext)
 var result: string;
 begin
-  result := FindMessage(Hash32(
+  //resourcestrings may be not in English
```
synopse: this comment (and below) is a bit confusing for sure -> need to delete old code, and fix comments

SQLite3/mORMoti18n.pas
```diff
@@ -2225,15 +2250,15 @@ function AddOnceDynArray(const S: WinAnsiString): integer;
...
+var buf: array[0..4095] of WideChar;
+    s: RawUTF8;
     i: PtrInt;
 begin
   result := true;
   if (PtrInt(lpszType)<>PtrInt(RT_STRING)) then exit;
   i := (PtrInt(lpszName)-1)shl 4;
   for i := i to i+15 do begin // resourcestrings are stored by groups of 16
-    SetString(s,buf,LoadStringA(hInstance,i,buf,sizeof(buf)));
+    SetString(s,buf,LoadStringW(hInstance,i,buf,sizeof(buf)));
```
synopse: this is IMHO wrong: s is UTF-8 while LoadStringW returns UTF-16 content

SynCommons.pas
```diff
@@ -11303,7 +11303,7 @@ function Hash32(Data: pointer; Len: integer): cardinal; overload;
 // - has less colision than Adler32 for short strings
 // - is faster than CRC32 or Adler32, since use DQWord (128 bytes) aligned read
 // - uses RawByteString for binary content hashing, whatever the codepage is
-function Hash32(const Text: RawByteString): cardinal; overload;
+function Hash32(const Text: RawUTF8): cardinal; overload;
```
synopse: not acceptable (it breaks existing user code), nor needed RawByteString as input allows to pass a RawUTF8

paustr:
I have an application, which default langage is polish and all it's text are in polish.
Now I need to provide language selection (english translation will not be needed) and we have to ba able to change langugage many times without restart an application.
I use SetCurrentLanguage method but it wasn't available unless {$undef USEFORMCREATEHOOK} was used (note that I dint't want to use LVCL directive, whitch undefines USEFORMCREATEHOOK).
I also made some changes in ExtractAllResources procedure to write correctly polish texts into .messages file.
I want to use mORMori18n bot in VCL and FMX(in FMX only for translating resourcestrings), so I had to use FMXAPP directive to make it compilable under FireMonkey.
Please look at sample application

synopse:
Please see my review above, to make some needed corrections in your branch, then try again the pull request.

SQLite3/mORMoti18n.pas
paustr: removed

SQLite3/mORMoti18n.pas
paustr: SetString method causes a problem unless buf is declared as array od WideChar. Please look at callstack
1. LoadStringA and buf is array of AnsiChar
cb_enumstringproc_1
1. LoadStringW and buf is array od WideChar
cb_enumstringproc_2

SynCommons.pas
paustr: Change is reverted.

paustr: I still need Hash32(const Text: RawUTF8), so it was written in mORMoti18n.
After SetCurrentLanguage(lngEnglish) method was called, all resourcestring need to be tranlated and method GetText was called. There is an implicit conversion from RawUTF8 to RawByteString and it causes loss of data. Please look at attached call stacks.
1. function Hash32(const Text: RawByteString) was used 

setcurrentlangugage_en

2. function Hash32(const Text: RawUTF8) was used
setcurrentlangugage_en2

synopse: Don't trust the debugger. IMHO this is a debugger issue/bug/feature, which makes on its side a  conversion when displaying the value in the IDE.
There is no actual conversion in the generated assembly when you assign a RawUTF8 to a RawByteString (use Alt-F2 to see it).

SQLite3/mORMoti18n.pas
synopse: if s is a RawUTF8 you should NOT use UTF-16 input
use `RawUnicodeToUtf8()` instead of `SetString()`

SynCommons.pas
paustr: But Hash32(const Text: RawByteString) called inside TLanguageFile.Translate(var English: string) returns result which is not equal to the hash saved in .messages file and so that TLanguage.FindMessage returns an empty string.
When I use Hash32(const Text: RawUTF8)  both for writing strings to .messages file (inside ExtractAllResources) and inside TLanguageFile.Translate method I'm getting the same results.

synopse:
Outdated.

Please reopen after resolution if needed.

PR #148, opened 2018-10-30
SyNode debugger improvements

pavelmash:
- use `TRawUTF8ListLocked` instead of `TRawUTF8ListHashedLocked` for messages queues 
- fix bug: detach debugger in case remote debugger client close socket without sending `detach` command (abnormal termination, network error, closing of FF window while standing on break-point)
- return empty response for "focus" command - nothing to focus in our case


pavelmash:
@ab - please, merge this into master

PR #150, opened 2018-11-04
Auxiliary RTree fields

esmondb:
strip '_' from auxiliary RTree field names


synopse:
If I remember correctly this was on purpose otherwise the table creation won't work as expected.

 esmondb <notifications@github.com> a écrit :

esmondb:
Yes, but I should have solved that with a new TSQLPropInfoAttribute of aAuxiliaryRTreeField.

PR #151, opened 2018-11-16
synode base64 decoder optimization

pavelmash:
# Fixed
 - Buffer.from(str, 'base64') will work correctly in case `str` is representing as 2-byte string inside SM engine (can happens in case `str` extracted from string with non-latin characters)

# Added
 - Buffer.from(str, 'base64') can decode base64 with wite-spaces inside (for example SOAP XMLs separate long base64 string by \r\n)
 
# Optimized
 - x5 speedup of Buffer.from(str, 'base64')


pavelmash:
@ab - please, merge this pull request to master (and to the fossil repo). Thanks

PR #152, opened 2018-11-17
unmapping memory mapped files for Unix

pavelmash:
Unix: Memory mapped files must be un-mapped for any non zero buffer size
Without this fix files remains mmaped until process exit


pavelmash:
@ab - this is critical fix.  Without is process can't recreate file if this file is touched by TMemoryMap even after call to TMemoryMap.UnMap

PR #153, opened 2018-11-23
SyNode: Embadding JavaScript files into executable as a resources

pavelmash:
- JavaScript files from `core_modules` folder now embedding  into the executable
- feature can be disabled by CORE_MODULES_IN_RES define in the `SyNode.inc`
- `core_res` tool to create res file for embedding  from the specified list of JS source


pavelmash:
@ab - feature verified and can be merged into master

PR #157, opened 2018-12-01
TextColor - prevent unnecessary syscalls

pavelmash:
For FPC+Linux every call to write() =  syscall write

Since every syscall is: 
 - atomic
 - expensive

after this patch console coloring become a little bit faster and atomic (calling TextColor from different thread do not brake a terminal commands on 2 part)

(verified using `strace`)

### DIFF
--- a/SynCommons.pas
+++ b/SynCommons.pas
@@ -55888,6 +55888,7 @@ begin
   if ord(color)>=8 then
     write(#27'[1;3',AnsiTbl[(ord(color) and 7)+1],'m') else
     write(#27'[0;3',AnsiTbl[(ord(color) and 7)+1],'m');
+  ioresult;
 end;
 {$I+}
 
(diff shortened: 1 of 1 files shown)


SynCommons.pas
```diff
@@ -55886,10 +55886,8 @@ procedure TextColor(Color: TConsoleColor);
     exit;
   TextAttr := ord(color);
   if ord(color)>=8 then
-    write(#27'[1;3') else
-    write(#27'[0;3');
-  write(AnsiTbl[(ord(color) and 7)+1],'m');
-  ioresult;
```
synopse: why did you remove ioresult call ? IIRC it doesn't call any sys API.

pavelmash: This is my mistake. I think call to ioresult it do nothing in this case, but now I understand it set global `InOutRes` variable to 0.

PR #159, opened 2018-12-03
Fix some valgrind warnings 

pavelmash:
This patch allow to run mormot based app under varglind - `$define HASAESNI` should be commented for this.

Also fixed potential issue in UnCamelCase (as reported by valgrind)


SynCommons.pas
```diff
@@ -37500,7 +37500,7 @@ function UnCamelCase(D, P: PUTF8Char): integer;
   Space := D;
   DBeg := D;
   SpaceBeg := D;
-  if (D<>nil) and (P<>nil) then // avoid GPF
+  if (D<>nil) and (P<>nil) then begin // avoid GPF
   repeat
```
synopse: please add properly the indentation spaces needed by the new begin..end block. ;)

pavelmash: I do not add indentation to simplify a review for a maintainer - if I will diff become non-readible :)

pavelmash: I'm add an indentation.  Actually I'm use gitlab in my every day work and miss a "hide whitespace changes" feature in github diff view

synopse: indentation allows direct merge - I don't have problem to do the proper review even if github doesn't handle it well, but I would like to be able to make a direct merge :)

PR #160, opened 2018-12-04
new functions to handle unquoting CSV strings

maciej-izak:


maciej-izak:
Done (with small fix) in https://github.com/synopse/mORMot/pull/160/commits/4b6e0a3d3f6ea642ffba507fc9ff7f7029bc03e8

pavelmash:
In  [CSV rfc4180](https://tools.ietf.org/html/rfc4180) double - quotes (") are used for strings. May be change default Quote char from single quoter to double in GetUnQuoteCSVItem ?

maciej-izak:
@pavelmash quote `'` is used because in RawUTF8ArrayToQuotedCSV the  `'` is also used but I agree that `"` is more correct.

PR #162, opened 2018-12-05
get a low level connection ID from reverse proxy server header

pavelmash:
TODO:
 - remove usage of SynCommons - temporary added for GetInt64
 - discuss changing of ConnectionID type to Int64 in all classes

### DIFF
--- a/SynCrtSock.pas
+++ b/SynCrtSock.pas
@@ -3156,7 +3156,6 @@ implementation
 
 {$ifdef FPC}
 uses
-  SynCommons, // MPV - temporary for GetInt64
   dynlibs;
 {$endif}
 
@@ -6228,7 +6227,7 @@ begin
   finally
     LeaveCriticalSection(fServer.fProcessCS);
   end;
-  if (aServerSock.RemoteConnectionID > 0) then
+  if (aServerSock.RemoteConnectionID <> 0) then
     fConnectionID := aServerSock.RemoteConnectionID
   else
     fConnectionID := fServer.NextConnectionID;
@@ -6583,6 +6582,7 @@ function THttpServerSocket.GetRequest(withBody: boolean=true): boolean;
 var P: PAnsiChar;
     i, L: integer;
     H: ^PAnsiChar;
+    pH: PAnsiChar;
     maxtix, status: cardinal;
     reason: SockString;
 begin
@@ -6622,7 +6622,8 @@ begin
         H := pointer(Headers);
         for i := 1 to length(Headers) do
         if IdemPChar(H^,pointer(fServer.fRemoteConnIDHeaderUpper)) and (H^[L]=':') then begin
-          fRemoteConnectionID := GetInt64(pointer(@H^[L+1]));
+          pH := H^+L+1; while (pH^=' ') do inc(pH);
+          fRemoteConnectionID := GetNextItemUInt64(pH);
           break;
         end else
           inc(H);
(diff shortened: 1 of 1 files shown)


SynCrtSock.pas
```diff
@@ -6209,7 +6228,10 @@ constructor THttpServerResp.Create(aServerSock: THttpServerSocket; aServer: THtt
   finally
     LeaveCriticalSection(fServer.fProcessCS);
   end;
-  fConnectionID := fServer.NextConnectionID;
+  if (aServerSock.RemoteConnectionID > 0) then
```
achechulin: I think should be:
`aServerSock.RemoteConnectionID` <> 0

SynCrtSock.pas
```diff
@@ -3144,6 +3156,7 @@ implementation
 
 {$ifdef FPC}
 uses
+  SynCommons, // MPV - temporary for GetInt64
```
synopse: you can use GetNextItemUInt64()

SynCrtSock.pas
synopse: +1

SynCrtSock.pas
```diff
@@ -6594,6 +6616,17 @@ function THttpServerSocket.GetRequest(withBody: boolean=true): boolean;
         end else
           inc(H);
       end;
+      // remote connection ID
+      L := length(fServer.fRemoteConnIDHeaderUpper);
+      if L<>0 then  begin
+        H := pointer(Headers);
+        for i := 1 to length(Headers) do
+        if IdemPChar(H^,pointer(fServer.fRemoteConnIDHeaderUpper)) and (H^[L]=':') then begin
```
synopse: why not just include the ':' in the fRemoteConnIDHeaderUpper ?

SynCrtSock.pas
```diff
@@ -8629,6 +8663,19 @@ procedure THttpApiServer.Execute;
...
         Context.fUseSSL := Req^.pSslInfo<>nil;
         Context.fInHeaders := RetrieveHeaders(Req^,fRemoteIPHeaderUpper,RemoteIP);
+        // compute remote connection ID
+        L := length(fRemoteConnIDHeaderUpper);
+        if L<>0 then begin
+          P := Req^.Headers.pUnknownHeaders;
+          if P<>nil then
+          for i := 1 to Req^.Headers.UnknownHeaderCount do
+            if (P^.NameLength=L) and IdemPChar(P^.pName,Pointer(fRemoteConnIDHeaderUpper)) then begin
+              SetString(RemoteConn,p^.pRawValue,p^.RawValueLength);
```
synopse: if you use GetNextItemUInt64(), local RemoteConn is not needed

SynCrtSock.pas
pavelmash: because we need  header without ':' in HTTP.SYS section

SynCrtSock.pas
pavelmash: We need local RemoteConn because `p^.pRawValue` not #0 ended (I adds a comment for this)

SynCrtSock.pas
```diff
@@ -1901,8 +1913,8 @@   THttpServer = class(THttpServerGeneric)
     // - expects the port to be specified as string, e.g. '1234'; you can
     // optionally specify a server address to bind to, e.g. '1.2.3.4:1234'
     // - you can specify a number of threads to be initialized to handle
-    // incoming connections (default is 32, which may be sufficient for most
-    // cases, maximum is 256) - if you set 0, the thread pool will be disabled
+    // incoming connections. Default is 32, which may be sufficient for most
```
pavelmash: I remove ( ) from comment because Lazarus parser not work correctly with ( ). Ctrl+Shift+Down not work (

PR #166, opened 2018-12-15
json serializer adds a T00:00Z to dates without time in UTC0

pavelmash:
In case woDateTimeWithZSuffix Option is passed to TJSONSerializer and TDateTime value do not contains a time then T00:00Z should be added to the resulting string. 

In other case FireFox and Internet Explorer based client fails to parse
```
new Date('2017-01-01Z')
```
Chromium will

### DIFF
--- a/SQLite3/mORMot.pas
+++ b/SQLite3/mORMot.pas
@@ -52244,7 +52244,9 @@ var Added: boolean;
               Add('"');
             AddDateTime(D64);
             if woDateTimeWithZSuffix in Options then
-              Add('Z');
+              if frac(D64)=0 then // FireFox can't decode short form "2017-01-01Z"
+                AddShort('T00:00Z') else
+                Add('Z');
             Add('"');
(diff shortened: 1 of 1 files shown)


SQLite3/mORMot.pas
```diff
@@ -52244,7 +52244,9 @@ procedure TJSONSerializer.WriteObject(Value: TObject; Options: TTextWriterWriteO
               Add('"');
             AddDateTime(D64);
             if woDateTimeWithZSuffix in Options then
-              Add('Z');
+              if frac(D64)=0 then // FireFox can't decode short form "2017-01-01Z"
```
synopse: won't it break some existing code?
perhaps worth a new option

pavelmash: I don't think it's break something - in case of SynCommon - based clients deserialization will work with new format. In case with JavaScript  based client this is mandatory changes

PR #169, opened 2018-12-31
FPC fix: incorrect usage of  TPrecisionTimer.ProfileCurrentMethod in SybDBOracle

pavelmash:
If result of `TPrecisionTimer.ProfileCurrentMethod` is not assigned to the local variable FPC destroy IUnknown immediately. 
So for FPC we should either create a local IUnknown var, or not use a .ProfileCurrentMethod at all. 
The second is better - in this case we do not create an unnecessary `TPrecisionTimerProfiler` class instance

### DIFF
--- a/SynCommons.pas
+++ b/SynCommons.pas
@@ -15879,6 +15879,7 @@ type
     // - typical use is to declare a fTimeElapsed: TPrecisionTimer protected
     // member, then call fTimeElapsed.ProfileCurrentMethod at the beginning of
     // all process expecting some timing, then log/save fTimeElapsed.Stop content
+    // FPC TIP: result should be assigned to the local variable of IUnknown type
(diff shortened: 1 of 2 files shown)


SynCommons.pas
```diff
@@ -15879,6 +15879,7 @@   {$ifdef UNICODE}TPrecisionTimer = record{$else}TPrecisionTimer = object{$endif
     // - typical use is to declare a fTimeElapsed: TPrecisionTimer protected
     // member, then call fTimeElapsed.ProfileCurrentMethod at the beginning of
     // all process expecting some timing, then log/save fTimeElapsed.Stop content
+    // FPC TIP: result should be assigned to the local variable of IUnknown type
```
synopse: typo
// - FPC TIP: ... to a local variable
(add a leading - and use "a" instead of "the")

PR #172, opened 2019-01-03
Memory leak in JS_Enumerate function fixed

ssoftpro:


synopse:
sorry I did the merge - I thought it was into the trunk branch! :(

pavelmash:
It's OK to be in synodeCleanup - we need to compile libsynmozjs after this patch (and upload new so/dll version). I will test this patch carefully and when we merge all synodeCleanup changes to master

PR #173, opened 2019-01-30
prevent sending Content-Type: !STATICFILE header

pavelmash:
This header shoud not be sent


pavelmash:
After today`s update of Google Chrome to v72 in case of two Content-Type header in response the resulting type that sees the web browser is unpredictable

PR #178, opened 2019-03-09
Update SynCrtSock.pas

devaex:
Don't compile under Delphi 10.3 Rio.
[dcc32 Error] SynCrtSock.pas(11489): E1012 Constant expression violates subrange bounds

PS: I know Curl should be used in Linux.
But I'm coding in Delphi and then generating the final version on Lazarus.

### DIFF
--- a/SynCrtSock.pas
+++ b/SynCrtSock.pas
@@ -11716,7 +11716,7 @@ begin
       on E: Exception do begin
         if curl.Module<>0 then
           FreeLibrary(curl.Module);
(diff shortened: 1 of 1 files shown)


synopse:
This fix won't work, since 0=nil indicates that libcurl has not been tested yet (a global variable equals 0 by default), whereas -1 was set to test libcurl presence only once.

synopse:
I guess https://github.com/synopse/mORMot/commit/4cb79e151cc2d036cd557f75fd1c4c7201a9b0c2 is the correct fix for this.

PR #181, opened 2019-03-21
fixed NewSQLHttpClient function

VitalyPopov:
https://synopse.info/forum/viewtopic.php?id=4494


VitalyPopov:
no need anymore, because NewSQLHttpClient is gone

PR #182, opened 2019-04-02
try to fix non-english exception logging for FPC

pavelmash:
this patch fix logging of non-english exception text on FPC. Example:
```
program Project1;
uses
  sysutils, SynLog;
begin
  TSynLog.Family.Level := LOG_VERBOSE;
  try
    raise Exception.create('ошибка');
  finally
  end;
end.
```

### DIFF
--- a/SynCommons.pas
+++ b/SynCommons.pas
@@ -54393,16 +54393,12 @@ begin
     AddNoJSONEscape(pointer(s),L); // identified as a BLOB content
     exit;
   end;
-  {$ifdef FPC}
-  CodePage := CP_UTF8;
-  {$else}
   if CodePage<0 then
     {$ifdef HASCODEPAGE}
     CodePage := StringCodePage(s);
     {$else}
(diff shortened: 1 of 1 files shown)


pavelmash:
In my case TSynAnsiConvert.Engine(0) return TSynAnsiFixedWitdh(1251) because 
```
CurrentAnsiConvert := TSynAnsiConvert.Engine(GetACP); // GetACP == 1251 
```

pavelmash:
I made another attempt to fix based on fact what for fpc > 3 string = utf8 string as [described in FPC WiKi](http://wiki.freepascal.org/Unicode_Support_in_Lazarus)

synopse:
So GetAcp should return utf8 I guess. Otherwise other functions could fail.

 pavelmash <notifications@github.com> a écrit :

pavelmash:
Closed as do not reproduced anymore (at last with fpc 3.2)

PR #185, opened 2019-04-09
Fix/http api content type empty body

pavelmash:
The same as in #184 but for HTTPApi based server


pavelmash:
sory, wrong parent for brunch

PR #186, opened 2019-04-09
HttpApi server: allow to send Content-Type for empty body

pavelmash:
The same as in #184 but for HTTP API based server


pavelmash:
Can this patch be merged? Or it broke something in mORMot?

PR #187, opened 2019-04-13
use FPC algorithm for UTF8List capacity grow

pavelmash:
My application widely use TRawUTF8List to store name-value pairs and minimum capacity of 256 is seams to be too large.

Both C++ vector and FPC TFPList use more flexible algorithms. I'm copy-paste FPC algorithm and test it with my real use-case (`/usr/bin/time myApp`)

Current algo:
```
Maximum resident set size (kbytes): 278840
Minor (reclaiming a frame) page faults: 81035
```
FPC algo:
```
Maximum resident set size (kbytes): 274920
Minor (reclaiming a frame) page faults: 79774
```

Memory allocation is better with FPC algorithm.
Benchmark performance is also increased.

### DIFF
--- a/SynCommons.pas
+++ b/SynCommons.pas
@@ -2184,6 +2184,7 @@ var strcspn: function (s,reject: pointer): integer = strcspnpas;
 
 {$ifndef ABSOLUTEPASCAL}
 {$ifdef CPUINTEL}
+{$ifdef HASAESNI}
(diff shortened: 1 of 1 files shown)


synopse:
I guess https://github.com/synopse/mORMot/commit/a84dfa7e5178a54e7795fc89e35f6e507ddf7e12 is a more systematic way of applying such a better array growing stategy.

PR #188, opened 2019-04-16
Introducing Runtime Packages

mdbs99:
Some comments:
1. All packages are compiling, except `mormot_zeos` - see item 3.
1. I haven't added `SynFPCCMemAligned`, `SynFPCLinux`, `SynFPCSock` on `mormot_base`, as these files are platform (Linux vs Windows) dependent.
1. `SynDBZeos` uses `{$I Zeos.inc}` but this file do not have direct access to this inc file without a hard-coded path. I don't know how to solve this yet.


synopse:
Without SynFPCLinux and SynFPCSock, SynCommons and SynCrtSock won't compile on Linux.

mdbs99:
SynCommons was included on mormot_base package.
Units that belongs to an specific platform, should be used in other units using `IFDEF`. They just need to exist on the PATH - already setup in all packages.

mdbs99:
This https://github.com/synopse/mORMot/pull/188/commits/a3a4197f3d5786961e1b5e0462dbb1efeca827d8 fixes item 3.

mdbs99:
@synopse do I need to do more improvements to merge? Please, let me know.

PR #189, opened 2019-04-21
SynLog: unicode environment variable in header

pavelmash:
change encoding of logged environment variables form OEM to UTF8, since all other content on log is in UTF8. 
In case we log unicode variables values in OEM encoding (as returned by GetEnvironmentStringsA) it's become unreadable on other computers, because we even don't know what code page is set as console code page on the log source machine

### DIFF
--- a/SyNode/SyNode.pas
+++ b/SyNode/SyNode.pas
@@ -81,8 +81,10 @@ unit SyNode;
 interface
 
 uses
-  {$ifndef FPC}
+  {$ifdef MSWindows}
   Windows,
+  {$endif}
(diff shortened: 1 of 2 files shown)


pavelmash:
Sample. 
``` set french=français ```
With this patch Environment variables block in log is:

```Environment variables=ALLUSERSPROFILE=C:\ProgramData ..... french=français ....```

Without this patch

```Environment variables=ALLUSERSPROFILE=C:\ProgramData ..... french=francais ....```

As we can see **ç** letter is replaced by **c**.

PR #193, opened 2019-04-29
logger fixes

AlekId:


SynLog.pas
```diff
@@ -2243,8 +2243,10 @@ class procedure TSynMapFile.Log(W: TTextWriter; aAddressAbsolute: PtrUInt;
     exit;
   {$ifdef FPC}
   s := BacktraceStrFunc(pointer(aAddressAbsolute));
-  if Pos('SynLog.pas',s)=0 then // don't log internal calls
+  if Pos('SynLog.pas',s)=0 then begin // don't log internal calls
     W.AddShort(s);
+    W.AddShort(#14);
```
synopse: what is this #14 for?

SynLog.pas
```diff
@@ -5342,7 +5346,7 @@ procedure TSynLogFile.ProcessOneLine(LineBeg, LineEnd: PUTF8Char);
   if (fCount<=fLineHeaderCountToIgnore) or (LineEnd-LineBeg<24) then
     exit;
   if fLineLevelOffset=0 then begin
-    if (fCount>50) or not (LineBeg[0] in ['0'..'9']) then
+    if  not (LineBeg[0] in ['0'..'9']) then
```
synopse: why is fCount>50 wrong?

SynLog.pas
```diff
@@ -5682,6 +5686,7 @@ procedure TSynLogFileView.AddInMemoryLine(const aNewLine: RawUTF8);
     tm: cardinal;
 begin
   tm := fThreadMax;
+  fLineLevelOffset:=0;
```
synopse: this will break the expected behavior of this method, for sure

SynLog.pas
AlekId: because it limits viewer to 50 events, that for sure.

SynLog.pas
AlekId: this is lifehack to break stacktrace in viewer -- I use it to break call stack to separate lines. Don''t know but #10 didnt worked.
ttt

SynLog.pas
AlekId: it's buggy, the multithreaded Lazarus app just don't able to send log to viewer. Or viewer can't read it

AlekId: that way it works, the other **dont** . Fix it better, if you can. 
My goal is to make logger functional for Lazarus app. I **made** it functional. I don't have time to delve into spaghetti, which implements logger logic.

SynLog.pas
synopse: it is expected to have the header available and discovered within the 50 first text files - this won't limit t 50 events; if it is not the case, then the log format is not correct

SynLog.pas
synopse: https://stackoverflow.com/help/mcve is always better than carping I guess

synopse:
No feedback since more than a year.
Closed.

PR #194, opened 2019-04-30
fix SIGSEGV in ToUTF8 function for x64 target

pavelmash:
for int64 buffer should be 24 char. Program bellow raise SIGSEGV for x64 with optimization level >=2 
```
program Project1;
uses
  SynCommons;
var
  t: TThreadID;
begin
  t := 140737295120128;
  writeln(ToUTF8(t));
end.
```

### DIFF
--- a/SynCommons.pas
+++ b/SynCommons.pas
@@ -24198,16 +24198,15 @@ end;
 
 {$endif DEFINED_INT32TOUTF8}
 
-{$ifndef CPU64} // already implemented by ToUTF8(Value: PtrInt) below
-function ToUTF8(Value: Int64): RawUTF8;
+{$ifdef CPU64}
+function ToUTF8(Value: PtrInt): RawUTF8;
 var tmp: array[0..23] of AnsiChar;
     P: PAnsiChar;
 begin
   P := StrInt64(@tmp[23],Value);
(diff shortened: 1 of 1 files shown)


SynCommons.pas
```diff
@@ -24209,12 +24209,21 @@ function ToUTF8(Value: Int64): RawUTF8;
...
 function ToUTF8(Value: PtrInt): RawUTF8;
+{$ifdef CPU64}
+var tmp: array[0..23] of AnsiChar;
+    P: PAnsiChar;
+begin
+  P := StrInt64(@tmp[23],Value);
+  FastSetString(result,P,@tmp[23]-P);
+end;
+{$else}
 var tmp: array[0..15] of AnsiChar;
```
synopse: replacing it with `var tmp: array[0..23] of AnsiChar;` sounds simplier thand a ifdef/endif since StrInt64=StrInt32

pavelmash: Let's keep StrInt64 where
 - call to StrInt32 inside Int64 convertion is confused
 - StrInt32 is implemented in assembler and it's very unclear what it can convert both Int32 and Int64

synopse:
latest commit above should fix your issue

pavelmash:
fixed by [97e3a50]

PR #197, opened 2019-05-09
fix LoadLibrary in SynGSSAPI &  SynWinSock for FPC x64

pavelmash:
Under FPC use TLibHandle instead of THandle. THandle is 32bit under x64 FPC
Using THandle cause a very strange exception what depends on how many libraries are loaded by operation system yet.

### DIFF
--- a/SynDBOracle.pas
+++ b/SynDBOracle.pas
@@ -1776,13 +1776,6 @@ begin
     result := OCI_WE8MSWIN1252;
 end;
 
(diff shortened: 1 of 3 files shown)


synopse:
What about the other occurences of LoadLibrary() in the framework?

pavelmash:
I check all of them. All other occurrences either use HMODULE (this is OK) or already use TLibHandle

PR #199, opened 2019-05-15
Proposal to allow GetAsDocVariant return ID instead of RowID

devaex:
And small change in RetrieveDocVariant to use this new param.

from 
Rec.GetAsDocVariant(True,bits,result,nil)

to
 Rec.GetAsDocVariant(True,bits,result,nil,IdemPropNameU('ID=?',FormatSQLWhere));


synopse:
Your pull request contains a very old version of the unit.
It is therefore not mergeable. I will do the fix directly.
See https://synopse.info/forum/viewtopic.php?pid=29583#p29583

devaex:
Thanks AB,

I had some problem due to the size of the Mormot.pas that went beyond the github limit.

I was without access to a git client, so I created a fork, deleted Mormot.pas and sent the updated one to my fork,

Believing that Github instead of comparing only the modifications, compared everything with much older modifications.

I see now that this was not a good decision.

I appreciate your attention.

devaex:
As a suggestion, it might be interesting to add replaceRowIDWithID to GetSimpleFieldsAsDocVariant as well.

Best Regards

PR #205, opened 2019-05-27
Destroy expired connections more quickly

e-martin:
Check that the connections do not grow indefinitely in situations where is needed create connections for posterior destroy.


e-martin:
It's perfect for me.

Thanks.

PR #208, opened 2019-05-31
SynCommons.pas Warning Duplicate constructor

OkanMS:
Hi,

I'm getting the following warning.

[dcc32 Warning] W1029 Duplicate constructor 'ESynException.CreateLastOSError' with identical parameters will be inacessible from C++


synopse:
I guess there is something wrong with your pull request: you want to merge 42 commits into feature/synodeCleanup from master, whereas it should be from the other way round...

PR #220, opened 2019-07-03
Fix/win http upload

ysair:


ysair:
SockString array index start at 1, but PByteArray array index start at 0.

this commit 
[https://github.com/synopse/mORMot/commit/210ff619dfad991b50ea8aa9136db3af341f848f](url) 
change aData(SockString) to @PByteArray(aData), Cause this bug

PR #222, opened 2019-07-17
Remove warnings compiling Packages

mdbs99:
There is an situation on mORMot Packages that is inconvenient.

If I add two packages (or more) like momort_base, mormot_rest, and mormot_sqlite3 I got messages like this:

```
Warning: other unit files search path (aka unit path) of "mormot_sqlite3 1.18" contains "<PATH>\lib", which belongs to package "mormot_base"
Warning: other unit files search path (aka unit path) of "mormot_rest 1.18" contains "<PATH>\lib\SQLite3", which belongs to package "mormot_sqlite3"
Warning: other unit files search path (aka unit path) of "mormot_base 1.18" contains "<PATH>\lib", which belongs to package "mormot_sqlite3"

Note: Duplicate unit "mORMotWrappers" in "mormot_base 1.18", orphaned ppu "<PATH>\lib\Packages\lib\i386-win32\mORMotWrappers.ppu"
Note: Duplicate unit "mORMotWrappers" in "mormot_rest 1.18", ppu="<PATH>\lib\Packages\lib\i386-win32\mORMotWrappers.ppu", source="<PATH>\lib\SQLite3\mORMotWrappers.pas"
```

This PR removes all warnings about packages compilation.

To do so, I needed to:
- all `mormot_ddd`, `mormot_ext`, `mormot_rest`, `mormot_sqlite3`, and `mormot_zeos` packages were merged into `mormot_base`, then they were removed
- special macro paths for original Zeos Packages, which was part of `mormot_zeos`, now is part of `mormot_base` — but it is used only when define `USEZEOS` in Custom Options
- introduced `ifdef USEZEOS` into `SynDBZeos` — if it isn't defined, `SynDBZeos` will be an empty unit
- renamed `mormot_crossplatform` to `mormot_cross`


Packages/mormot_base.lpk
```diff
@@ -16,8 +16,10 @@
         <UnitOutputDirectory Value="lib\$(TargetCPU)-$(TargetOS)"/>
       </SearchPaths>
       <Other>
-        <OtherDefines Count="1">
-          <Define0 Value="USEZEOS"/>
+        <CustomOptions Value="-dHASPACKAGES"/>
+        <OtherDefines Count="2">
+          <Define0 Value="HASPACKAGES"/>
```
mdbs99: @synopse I've chose `HASPACKAGES` because it already exists a `USEPACKAGES` (disabled by default) in `Synopse.inc`

PR #224, opened 2019-07-20
Disable dynamic Conditionals for ZeosLib

mdbs99:
Let's disable conditionals until next Lazarus version or "forever".
- see [comments](https://lists.lazarus-ide.org/pipermail/lazarus/2019-July/236729.html) and [patch](https://svn.freepascal.org/cgi-bin/viewvc.cgi/trunk/packager/interpkgconflictfiles.pas?root=lazarus&r1=61607&r2=61606&pathrev=61607) about it.

### DIFF
--- a/Packages/mormot_base.lpk
+++ b/Packages/mormot_base.lpk
@@ -21,6 +21,7 @@ begin
   IncPath += ';$(ProjOutDir);$PkgDir(zcore)\..\..\src';
 end;"/>
       <Other>
+        <CustomOptions Value="-dNOSYNDBZEOS"/>
         <OtherDefines Count="1">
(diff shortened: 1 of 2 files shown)


Packages/mormot_base.lpk
```diff
@@ -9,17 +9,11 @@
       <Version Value="11"/>
       <PathDelim Value="\"/>
       <SearchPaths>
-        <IncludeFiles Value="..;..\SQLite3"/>
+        <IncludeFiles Value="$PkgDir(zcore)\..\..\src;..;..\SQLite3"/>
```
mdbs99: We need this path because `SynDBZeos` uses `Zeos.inc`. It's not enough just add `zcomponent` into `mormot_base`.

PR #226, opened 2019-07-28
 Fixed: In the 'backupSuccess' step of a TSQLDatabaseBackupEvent event handler, you couldn't delete the target database file

edwinyzh:
Note: I needed to delete the target db file after after I've compress the db file into a .zip file


edwinyzh:
Hi ab, would you merge this fix? Thanks.

SynSQLite3.pas
```diff
@@ -3290,6 +3291,7 @@   TSQLDatabaseBackupThread = class(TThread)
     property DestDB: TSQLDatabase read fDestDB;
     /// the raised exception in case of backupFailure notification
     property FailureError: Exception read fError;
+    property BackupDestFile: RawUTF8 read fBackupDestFile;
```
synopse: please add a /// documentation comment

SynSQLite3.pas
```diff
@@ -5776,12 +5779,17 @@ procedure TSQLDatabaseBackupThread.Execute;
           raise ESQLite3Exception.Create('Backup process forced to terminate');
         SleepHiRes(fStepSleepMS);
       until false;
+
```
synopse: please don't add void lines to the source ;)

SynSQLite3.pas
edwinyzh: comment added

PR #227, opened 2019-07-28
Fixed a bug in TSChannelClient.Receive method. 

transmogrifix:
Data in SECBUFFER_EXTRA was not handled properly, it was appended to the output data buffer before being decrypted.

I have noticed this bug while trying to use web sockets (useBidirSocket and WebSocketsEnable) with a server behind NGINX reverse proxy with SSL.

### DIFF
--- a/SynWinSock.pas
+++ b/SynWinSock.pas
@@ -938,6 +938,7 @@ const
   SEC_E_OK = 0;
   SEC_I_CONTINUE_NEEDED = $00090312;
   SEC_I_RENEGOTIATE = $00090321;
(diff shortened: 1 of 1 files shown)


synopse:
I merged it, but something concerns me: your patch calls DecryptInput twice, even if InputCount is still 0 (no EXTRA bytes). @transmogrifix what do you think? I guess the following should be written:
```
  if InputCount = 0 then break;
  res := DecryptInput;
until false;
```

transmogrifix:
You are right, I missed that.
Thanks!

PR #228, opened 2019-07-31
Fix for "null" string value for nullable string columns

maciej-izak:
Fix for "null" string value received from SQL queries for nullable strings columns. "null" value was set as Null not as 'null'!

### DIFF
--- a/SQLite3/mORMot.pas
+++ b/SQLite3/mORMot.pas
@@ -31953,7 +31953,7 @@ begin
       with fTableMap[f] do
         if DestField=nil then
           SetID(aTableRow[TableIndex],Dest.fID) else
-          DestField.SetValue(Dest,aTableRow[TableIndex],false);
+          DestField.SetValue(Dest,aTableRow[TableIndex],TableIndex in fTable.fFieldParsedAsString);
 end;
 
 procedure TSQLRecordFill.Fill(aTableRow: PPUtf8CharArray; aDest: TSQLRecord);
@@ -31964,7 +31964,7 @@ begin
     with fTableMap[f] do
       if DestField=nil then
         SetID(aTableRow[TableIndex],aDest.fID) else
-        DestField.SetValue(aDest,aTableRow[TableIndex],false);
+        DestField.SetValue(aDest,aTableRow[TableIndex],TableIndex in fTable.fFieldParsedAsString);
 end;
 
 procedure TSQLRecordFill.ComputeSetUpdatedFieldBits(Props: TSQLRecordProperties;


synopse:
null should be received as 'null' in JSON, not as '"null"'

maciej-izak:
Probably you misunderstand my commit description. In JSON all is fine, the TSQLRecordFill.Fill is buggy.  Now is impossible to get Delphi 'null' string for nullable utf8 colum. Whatever is stored in JSON (I mean null or "null") the Delphis final value of column is set as Variant Null value (for both null and "null").

maciej-izak:
@synopse  With using rec.FillOne for the nullable column is  impossible to get proper 'null' string, the value for the field is equal to variant Null (expected is 'null').

Example to reproduce:
```pascal
{$APPTYPE CONSOLE}

uses
  SynCommons,
  SynDB,
  SynSQLite3,
  SynSQLite3Static,
  SynDBSQLite3,
  mORMot,
  mORMotDB,
  mORMotSQLite3
  ;

var
  Database: TSQLRestServerDB;
  Model: TSQLModel;
  Props: TSQLDBSQLite3ConnectionProperties;

type
  TSQLSampleRecord = class(TSQLRecord)
  private
    fFoo: TNullableUTF8Text;
  published
    property Foo: TNullableUTF8Text read fFoo write fFoo;
  end;

procedure InitializeApp;
begin
  Props := TSQLDBSQLite3ConnectionProperties.Create(SQLITE_MEMORY_DATABASE_NAME,'','','');
  Model := TSQLModel.Create([TSQLSampleRecord]);
  VirtualTableExternalRegisterAll(Model,Props,[]);
  Database := TSQLRestServerDB.Create(Model, SQLITE_MEMORY_DATABASE_NAME);
  Database.CreateMissingTables;
end;

procedure FinalizeApp;
begin
  Database.Free;
  Model.Free;
  Props.Free;
end;

function CreateRecord(const aFooValue: TNullableUTF8Text): TID;
var
  r: TSQLSampleRecord;
begin
  with TAutoFree.One(r, TSQLSampleRecord.Create) do begin
    r.Foo := aFooValue;
    result := Database.Add(r, true);
  end;
end;

procedure CheckRecord(id: TID);
var
  r: TSQLSampleRecord;
begin
  with TAutoFree.One(r, TSQLSampleRecord.CreateAndFillPrepare(Database, 'id = ?', [id])) do begin
    if r.FillOne then
      if NullableUTF8TextIsEmptyOrNull(r.Foo) then
        WriteLn('Foo is empty or null')
      else
        WriteLn('Foo = ''', NullableUTF8TextToValue(r.Foo), '''');
  end;
end;

var
  nullId: TID;
  strId: TID;
begin
  InitializeApp;

  nullId := CreateRecord(NullableUTF8TextNull);
  strId := CreateRecord(NullableUTF8Text('null'));
  CheckRecord(nullId);
  CheckRecord(strId);

  FinalizeApp;
  readln;
end.
```

current output:

```
  Foo is empty or null
  Foo is empty or null 
```

expected:
```
  Foo is empty or null
  Foo = 'null'
```

PR #230, opened 2019-08-14
SynGdiPlus: add support for image attributes

dhewg:
This adds the flat ImageAttributes GDI+ functions to TGDIPlus and
introduces the helper class TImageAttributes to work with image
attributes (which works like MS's ImageAttributes C++ class).

Finally, TSynPicture gained an overloaded Draw() procedure which
accepts a TImageAttributes argument. TGDIPlusFull gained an overloaded
DrawAntiAliased() in the same way. Both use DrawImageRectRect() internally,
so they expose a destination rect as well.

### DIFF
--- a/SynGdiPlus.pas
+++ b/SynGdiPlus.pas
@@ -255,7 +255,88 @@ type
   end;
   PGdipBitmapData = ^TGdipBitmapData;
 
+  TGpipImageAttributes = Pointer;
+
+  TColorAdjustType = (
+    ColorAdjustTypeDefault,
+    ColorAdjustTypeBitmap,
+    ColorAdjustTypeBrush,
+    ColorAdjustTypePen,
+    ColorAdjustTypeText,
+    ColorAdjustTypeCount,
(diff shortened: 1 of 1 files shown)


dhewg:
I'm using this with [QColorMatrix](https://www.codeguru.com/cpp/g-m/gdi/gdi/article.php/c3667/Playing-with-ColorMatrix.htm), which I ported to Delphi [QColorMatrix.txt](https://github.com/synopse/mORMot/files/3501358/QColorMatrix.txt). For example creating a disabled glyph:
```
q:= QColorMatrix.Create;
q.ScaleOpacity(0.5);
q.SetSaturation(0);
ia:= TImageAttributes.Create;
ia.SetColorMatrix(cm.ToColorMatrix);
sp.Draw(b.Canvas, rdst, rsrc, ia);
ia.Free;
q.Free;
```
If there's interest, I could add that as well, maybe as `TSynColorMatrix`?

dhewg:
ping

PR #234, opened 2019-09-12
TBidirServer test cases extended

X-FireFighter-X:
The extended test case "TestRestParameterSequence2" shows an error with special parameter sequence.


synopse:
You need to register your dynamic array for proper JSON serialization.

PR #238, opened 2019-09-30
Ignore sqlite3.obj in Git project

eugeneilyin:
Since SQLite3 .obj file are not part of mORMot package the sqlite3.obj file should be
added to .gitignore.

Add sqlite3.o is not necessary because *.o mask is already there.


synopse:
sqlite3.obj is needed for Delphi 32-bit, which is the main target of this github repository until now ...

eugeneilyin:
No, all functionality remain the same. It will not break anything. 

I'm not suggest to remove `sqlite3.obj` from mORMot framework.
This modification is just exclude this file from commits, uploads, etc. of GitHub project, because both files `sqlite3.obj` and `sqlite3.o` are [provided separately](https://synopse.info/files/sqlite3obj.7z) from the framework.

Usually these files unpacked into root folder. As the result GitHub add `sqlite3.obj` to repository and all future commits. This is uncomfortable and `sqlite3.obj` be add to commit unintentionally.
`.gitignore` prevents such cases and remove this file from GitHub interests.

Btw, `sqlite3.o` file is already excluded with `*.o` template in [line #46](https://github.com/synopse/mORMot/blob/726bfc0131f6019160d406fe2909adcf8d123e4d/.gitignore#L46) of `.gitignore` file.
All I asked with this is to be consistent and add `sqlite3.obj` as well.

eugeneilyin:
Both sqlite3.obj and sqlite3.o files already present in GitHub repository in /SQLite3 directory.
This pull request is not necessary.

PR #239, opened 2019-10-03
fixes for SMS

maciej-izak:
During using SMS compiler I found some non standard unexpected problems, for example `Assigned` should be used instead of `<>nil` because the second don't work for `unassigned` and some other unexpected events like releasing twice (or more) the same object... Also in template is used `Length` instead of `length` (this is big problem for SMS Variants - this part of template was nonfunctional)

This patch should improve stability.

### DIFF
--- a/CrossPlatform/SynCrossPlatformREST.pas
+++ b/CrossPlatform/SynCrossPlatformREST.pas
@@ -1673,7 +1673,7 @@ constructor TSQLRecord.Create(aClient: TSQLRest; aID: TID;
   ForUpdate: boolean=false);
 begin
   Create;
-  if aClient<>nil then
+  if Assigned(aClient) then
     aClient.Retrieve(aID,self,ForUpdate);
 end;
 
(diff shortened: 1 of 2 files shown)


CrossPlatform/SynCrossPlatformREST.pas
```diff
@@ -2974,8 +2974,9 @@ procedure TSQLRestClientURI.SetAsynch(var Call: TSQLRestURIParams;
             onError(self);
       if fAsynchCount>0 then
         dec(fAsynchCount);
-      if fAsynchCount=0 then
-        CallAsynchText; // send any pending asynchronous task
+      if fAsynchCount=0 then 
+        if Assigned(self) then
```
synopse: if self is void, then fAsynchCount access in the above line would fail - so what is it about?

maciej-izak: Good point. The main problem is that the fAsynchCount works and don't want to fail but at the same time CallAsynchText fails. I need to find some alternative to SMS... This is just hot-fix, I was not able to find the root of problem (and don't have much motivation for this - only root of DWScript is open source but whole JS part is not usable without commercial RTL).

synopse:
SMS seems (sadly) a dead project.
We will probably focus on TMS Web Core as alternative.

You can re-open this PR if you really need this.

PR #240, opened 2019-10-04
Feature Hide empty SynLog "Environment variables="

eugeneilyin:
If the `TSynLogFamily.NoEnvironmentVariable` set to `True` there is no need to add line
```
Environment variables=
```
without any content to each log header.


synopse:
OK - I have done the modification on my side, which was correct.

PR #247, opened 2019-10-25
Changes for ARM

LongDirtyAnimAlf:
Just a start. Not yet perfect.


SynCommons.pas
```diff
@@ -21046,11 +21052,11 @@ function GetEnumInfo(aTypeInfo: pointer; out MaxValue: Integer): PShortString;
     info := GetTypeInfo(aTypeInfo);
     base := info^.{$ifdef FPC_ENUMHASINNER}inner.{$endif}EnumBaseType;
     {$ifdef FPC} // no redirection if aTypeInfo is already the base type
-    if (base<>nil) and (base<>aTypeInfo) then
+    if (base<>nil) and (base^<>aTypeInfo) then
```
synopse: shouldn't it be `base{$ifndef HASDIRECTTYPEINFO}^{$endif}<>aTypeInfo` ?

SQLite3/mORMot.pas
```diff
@@ -21035,7 +21035,7 @@ function Deref(Info: PPTypeInfo): PTypeInfo;
 
 function GetTypeData(const info: TTypeInfo): pointer; {$ifdef HASINLINE}inline;{$endif}
 begin
-  result := AlignTypeData(PAnsiChar(@info.Name[1])+ord(info.Name[0]));
+  result := AlignTypeData(pointer(@info)+2+PByte(pointer(@info)+1)^);
```
synopse: I suspect `PByte(@info)`  could be Delphi compatible or perhaps `PAnsiChar(@info)+2+PByte(PAnsiChar(@info)+1)^`

synopse: but I don't see any logical difference with the former `PAnsiChar(@info.Name[1])+ord(info.Name[0])` expression, which is used in several places of the unit

LongDirtyAnimAlf:
shouldn't it be base{$ifndef HASDIRECTTYPEINFO}^{$endif}<>aTypeInfo ?
Yes !

The remainder of your remarks:
At the moment, ARM alignment causes me many (more) problems. Problems that are even dependent of the level of -O# settings.

So, I will continue my work on ARM.

I was hoping that the more trivial parts of my pull would be accepted and point you towards more undetected, but logical issues, unseen by me.

I will keep you informed.

LongDirtyAnimAlf:
Sidenote:

I am using ncgrtti.pas as a source to check the alignment requirements.

See:
procedure enumdef_rtti(def: tenumdef);
procedure setdef_rtti(def:tsetdef);

SQLite3/mORMot.pas
```diff
@@ -2764,6 +2764,8 @@   TInterfaceTypeData = record
   // & type NewType = OldType;
   // - user types defined as new types have this type information:
   // & type NewType = type OldType;
+  {$push}
```
synopse: this won't compile with Delphi I am afraid

LongDirtyAnimAlf:
I am closing this pull-request.
The work on ARM is still not finished. Especially the SOA has problems.
I am now also encountering a compiler bug, that must be circumvented.

Will make a new pull if the changes are finished.

PR #252, opened 2019-12-12
Pull for FPC trunk

LongDirtyAnimAlf:
Hi Ab,

As there are many improvements in FPC trunk , this final change allow compilation of the mORMot with trunk.

Its cheap, but working.

Please review.

Its about:
https://github.com/synopse/mORMot/pull/252/commits/fa32af49981e8ab481eefc351c3278334633e3f3#diff-83904bba45e1fdbc6476a49b32ba43bd


SynMongoDB.pas
```diff
@@ -4100,6 +4100,23 @@ function TBSONVariant.TryJSONToVariant(var JSON: PUTF8Char;
     end;
     result := true;
   end;
+  procedure Return(kind: shortint; P: PUTF8Char; GotoEndOfObject: AnsiChar='}');overload;
```
synopse: I got the idea.
This is IMHO a regression of trunk, since the previous code is fine with Delphi, since its early days.
Some small refactoring after merge.

synopse:
Indeed: the patch doesn't compile with Delphi.

LongDirtyAnimAlf:
That's a pity.
Some {$ifdef FPC} are needed in that case.
(I am no longer using Delphi at the moment, so I did not test it).
Sorry about this. it seemed good enough code.

PR #254, opened 2019-12-13
Fix compilation in Linux

cutec-chris:


synopse:
Those units are Windows-specific, so should NOT appear in the `uses` clause without proper $ifdef/$endif.

cutec-chris:
Thats simply only true for Delphi :)
Using the Lazarus Package, this units must be compiled. The alternative would be to make 2 separate packages for Windows and Linux but that would be also not good for cross plattform compatibility.

cutec-chris:
I have also Problems to get the actual mormot git working in Linux-64 with fpc 2.3.0_fixes it crashs always on line 49770:syncommons.pas Even when i use very simple examples like Sample 04
Would hope you could help to show me an fpc version/revision or an git commit id that could work.

PR #257, opened 2019-12-15
Use custom objarray writer for remote connection

LongDirtyAnimAlf:
I did encounter a bugger when using the object array custom readers and writer for a remote connection.

The reader of the remote result used the custom reader, but the writer did not.

This pull solves.

Please review.


synopse:
I don't think the initial code was correct, after deeper review.
If you use define a regular ObjArray (with WriteObject standard property-based serialization), HasCustomJSONParser returns TRUE so the previous code, with proper class instances indentation and optional comment is never used.

LongDirtyAnimAlf:
I do not know if I understand you correct.

Lets define a standard TJSONSerializer.RegisterObjArrayForJSON for a certain OjArray.
But breakpoint at the entrance of the DefaultCustom-Writer and -Reader.
If I put a breakpoint at (in mORMot.pas):
if method^.ArgsResultIsServiceCustomAnswer then

On my system, the Writer is not called when stopping at the invoke.
When continuing, the Reader does get called.

Again, I do not know if I understood your remark ... ;-)

LongDirtyAnimAlf:
Well, never mind, your latest commit works as expected !
Thanks for solving this !!

synopse:
Yes, https://github.com/synopse/mORMot/commit/a56b80f7d0dc7a04f3fd9940e82d36ba0f8d5192 is a rewrite of your commit, which didn't break the special serialization of T*ObjArrays (about indentation and optional comments for published properties).

PR #262, opened 2020-01-08
Check for Quoted Fields before check for Keywords

cutec-chris:
In actual implementation Keywords are not Quoted even when rpmQuoteFieldName is set. 
Ive reordered the Code so that check for rpmQuoteFieldName is done before check for Keyword (Keyword warning dosend makes sense too when Fieldname is Quoted)


cutec-chris:
Sorry for that trouble, but i have no idea how i can have it in my repo (since mormot dont work with fpc 3.3.1 without defining it, please double check)
and send you an merge request without it. Maybe i can make an separate branch for it or something. Ive check. 

But i would find it nice from you if you could double check it with fpc 3.3.1 (its not soo importend since it will wirk in fpc 3.4 and with fpc 3.2 it works too, but maybe some users use the trunk version)

PR #270, opened 2020-02-03
Fix slot.lastWSAError 

f-vicente:
Last changes in the code have hidden the error when the client does not close gracefully, (at least on Windows) lastWSAError always returns 0, because no operation has been performed on the socket. 

Calling Recv will return 0 (for gracefully closed connections) or an error which will be retrieved by GetWSALastError.

I'm not sure if this solution is the correct way of doing it, but works, and I don't think that calling Recv there will hurt.

### DIFF
--- a/SynCrtSock.pas
+++ b/SynCrtSock.pas
@@ -12552,6 +12552,14 @@ begin
 end;
 
 function TPollAsynchSockets.Stop(connection: TObject): boolean;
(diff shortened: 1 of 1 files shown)


synopse:
https://github.com/synopse/mORMot/commit/49d05bc00c24127ed24f97f4d76a81a5c71b981d should behave as expected.

PR #278, opened 2020-03-06
Optimized server socket acception on Linux

pavelmash:


pavelmash:
See https://synopse.info/forum/viewtopic.php?pid=31276 for details

PR #280, opened 2020-03-08
use TFP[Object]List instead of T[Object]List

pavelmash:
we do not need TList notification feature


synopse:
Perhaps using a dynamic array may be even better...

synopse:
I guess https://github.com/synopse/mORMot/commit/40750ba331e2a6f01b10b1586489beee5bc1c915 includes it.

PR #284, opened 2020-03-14
dev/urandom can return up to 256 bytes (2048 bit)

pavelmash:


synopse:
I am sorry, but https://linux.die.net/man/4/urandom states

> While some safety margin above that minimum is reasonable, as a guard against flaws in the CPRNG algorithm, no cryptographic primitive available today can hope to promise more than 256 bits of security, so if any program reads more than 256 bits (32 bytes) from the kernel random pool per invocation, or per reasonable reseed interval (not less than one minute), that should be taken as a sign that its cryptography is not skillfully implemented.

So the number of bytes is 32, not 256. You are confusing bytes and bits.

pavelmash:
I double check before this pull request. [Actual man page](http://man7.org/linux/man-pages/man4/random.4.html) (content is the same as in `man urandom`, source you provide is outdated) says:

       The O_NONBLOCK flag has no effect when opening /dev/urandom.  When
       calling read(2) for the device /dev/urandom, reads of up to 256 bytes
       will return as many bytes as are requested and will not be
       interrupted by a signal handler.  Reads with a buffer over this limit
       may return less than the requested number of bytes or fail with the
       error EINTR, if interrupted by a signal handler.

       Since Linux 3.16, a read(2) from /dev/urandom will return at most
       32 MB.  A read(2) from /dev/random will return at most 512 bytes (340
       bytes on Linux kernels before version 2.6.12).

So we actually can read 32MB but single read operation guaranty to return 256 bytes (2048 bits).
The random generator implementation in Linux is changed over years - see [good WiKi article](https://en.wikipedia.org/wiki//dev/random)

synopse:
The man page is not outdated. It is more accurate about how to *use* /dev/urandom for a PRNG.
There is no need to retrieve more than 256 bits for a PRNG.

Your link states something else. It states that you *can* get up to 256 bytes from /dev/urandom without being truncated - not that you *should* get 256 bytes.

Anyway, since our code fallbacks to /dev/random if /dev/urandom is not available (e.g. on BSD or old/buggy/hacked kernels), having a limitation to 32 bytes is a good idea. Your pull request could make GetEntropy() block if you try to read 256 bytes from /dev/random.

pavelmash:
Ok. Thanks for explanation.

PR #286, opened 2020-03-17
Good pull this time for BSD.

LongDirtyAnimAlf:
Please review.
All changes are necessary to get all greens on all systems.
mORMot100Arm64
mORMot100Darwin64
mORMot100FreeBSD
mORMot100OLinuxARM
mORMot100OLinuxARM64
mORMot100OpenBSD


LongDirtyAnimAlf:
About OpenBSD.
This distro has memory protection (w^x). That causes the small memory leak: patching code by the mORMot itself.
Its nothing to worry about, because it cannot grow.

SynSQLite3.pas
```diff
@@ -3609,6 +3609,7 @@ procedure InternalRank(Context: TSQLite3FunctionContext;
 procedure sqlite3InternalFree(p: pointer); cdecl;
 begin
   Freemem(p);
+  p:=nil;
```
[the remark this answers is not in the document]
LongDirtyAnimAlf: I think its a left-over from testing.
But, also saw a bugger on FPC list about some locals not set to nil after freeing.
I guess my thoughts were: better safe than sorry.

SynCommons.pas
```diff
@@ -16763,14 +16769,49 @@ implementation
...
+  This has been a very long standing issue on all arm/arm64 and BSD systems.
+  The sockets got corrupted by the variants test, leading to TestSQL3 failures.
+  Been hunting this for more than 2 years ...
+  It should/ could be investigated in more detail.
+  Also due to the fact that, for FPC this VTYPE_STATIC is often disabled in the source by defines !
+  *)
 var p: pointer; // more efficient generated asm with an explicit temp variable
 begin
   p := @v;
   if integer(PVarData(p)^.VType) and VTYPE_STATIC=0 then
```
synopse: Did you try to get rid of integer() here ?
for instance, with:
_if word(PVarData(p)^.VType) and VTYPE_STATIC = 0 then_

Perhaps it may be a nicer fix.

LongDirtyAnimAlf:
Thanks.
My major worry about mORMot is the varclear issue that has been "solved" by disabling.
It caused severe memory corruption. And that should never (and could not) happen.
It's a riddle for me, but I am very happy that this issue is gone.
Memory corruption is dangerous.

SynCommons.pas
LongDirtyAnimAlf: See my comment below.
I have no clue about this memory corruption. It should not happen. Its a compiler thing I guess.
A mystery for me. And I am not going to dive into the compiler.

PR #288, opened 2020-03-18
AES CTR mode is broken

karwos:
Hello.

The AES CTR mode is broken.

According to NIST SP 800-38A document, test vectors for AES128 CTR mode are:

> F.5.2 CTR-AES128.Decrypt
> Key 2b7e151628aed2a6abf7158809cf4f3c 
> 
> Init. Counter f0f1f2f3f4f5f6f7f8f9fafbfcfdfeff 
> 
> Block #1 
> 
> Input Block f0f1f2f3f4f5f6f7f8f9fafbfcfdfeff 
> 
> Output Block ec8cdf7398607cb0f2d21675ea9ea1e4 
> 
> Ciphertext 874d6191b620e3261bef6864990db6ce 
> 
> Plaintext 6bc1bee22e409f96e93d7e117393172a 
> 
> Block #2 
> 
> Input Block f0f1f2f3f4f5f6f7f8f9fafbfcfdff00 
> 
> Output Block 362b7c3c6773516318a077d7fc5073ae 
> 
> Ciphertext 9806f66b7970fdff8617187bb9fffdff 
> 
> Plaintext ae2d8a571e03ac9c9eb76fac45af8e51 
> 
> Block #3 
> 
> Input Block f0f1f2f3f4f5f6f7f8f9fafbfcfdff01 
> 
> Output Block 6a2cc3787889374fbeb4c81b17ba6c44 
> 
> Ciphertext 5ae4df3edbd5d35e5b4f09020db03eab 
> 
> Plaintext 30c81c46a35ce411e5fbc1191a0a52ef 
> 
> Block #4 
> 
> Input Block f0f1f2f3f4f5f6f7f8f9fafbfcfdff02 
> 
> Output Block e89c399ff0f198c6d40a31db156cabfe 
> 
> Ciphertext 1e031dda2fbe03d1792170a0f3009cee 
> 
> Plaintext f69f2445df4f9b17ad2b417be66c3710 

So decrypting a ciphertext: **874d6191b620e3261bef6864990db6ce9806f66b7970fdff8617187bb9fffdff5ae4df3edbd5d35e5b4f09020db03eab1e031dda2fbe03d1792170a0f3009cee**

Should yield to **6bc1bee22e409f96e93d7e117393172aae2d8a571e03ac9c9eb76fac45af8e5130c81c46a35ce411e5fbc1191a0a52eff69f2445df4f9b17ad2b417be66c3710** plaintext.

With current SynCrypto.pas following plaintext is generated:

**6bc1bee22e409f96e93d7e117393172a_2fd49f200e06fee142bb33586f61a3b7e0f54cbac8d2368ed0004e8786feebb2dfc95079569c95d480537e2279ab9e4c_**

The corruption yields after first AES block, and this is because wrong increasing offset in syncrypto.

Following can be verified by code snippet:

> var
>   A: TAesCtr;
>   AES_IV: TAESBLOCK;
>   Key,Iv,buf,buf2: array of byte;
> begin
> PFU_HexToDynBuffer('2b7e151628aed2a6abf7158809cf4f3c', @Key);
> PFU_HexToDynBuffer('f0f1f2f3f4f5f6f7f8f9fafbfcfdfeff', @Iv);
> PFU_HexToDynBuffer('874d6191b620e3261bef6864990db6ce9806f66b7970fdff8617187bb9fffdff5ae4df3edbd5d35e5b4f09020db03eab1e031dda2fbe03d1792170a0f3009cee', @buf);
> A := TAesCtr.Create(key[0], length(key)*8);
> Move(IV[0], AES_IV[0], length(AES_IV));
> A.IV:=AES_IV;
> A.Decrypt(@buf[0], @buf[0], Length(Buf));
> Memo1.Lines.Add(LowerCase(PFU_DynBufferToHex(buf)));   

Perhaps I haven't missed anything in Initialization code or Endianess settings? Rather no - data is referenced byte-wise rather than dword-wise.

### DIFF
--- a/SynCrypto.pas
+++ b/SynCrypto.pas
@@ -13054,9 +13054,9 @@ begin
   inherited; // CV := IV + set fIn,fOut
   for i := 1 to Count shr 4 do begin
     TAESContext(AES.Context).DoBlock(AES.Context,fCV,tmp);
-    inc(fCV[7]); // counter is in the lower 64 bits, nonce in the upper 64 bits
-    if fCV[7]=0 then begin // manual big-endian increment
-      j := 6;
+    inc(fCV[15]); // counter is in the lower 64 bits, nonce in the upper 64 bits
+    if fCV[15]=0 then begin // manual big-endian increment
+      j := 14;
       repeat
         inc(fCV[j]);
         if (fCV[j]<>0) or (j=0) then


SynCrypto.pas
```diff
@@ -13054,9 +13054,9 @@ procedure TAESCTR.Encrypt(BufIn, BufOut: pointer; Count: cardinal);
   inherited; // CV := IV + set fIn,fOut
   for i := 1 to Count shr 4 do begin
     TAESContext(AES.Context).DoBlock(AES.Context,fCV,tmp);
-    inc(fCV[7]); // counter is in the lower 64 bits, nonce in the upper 64 bits
-    if fCV[7]=0 then begin // manual big-endian increment
-      j := 6;
+    inc(fCV[15]); // counter is in the lower 64 bits, nonce in the upper 64 bits
```
synopse: https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Counter_(CTR) states that we should be "storing the nonce in the upper 64 bits and the counter in the lower 64 bits of a 128-bit counter block". This is what the current implementation does - and commented as such in this line.

Your modification uses the IV not as 64-bit nonce, but as full 128-bit CTR initialization.
This is another algorithm.

The problem is that if we merge this pull request, existing code using TAESCTR will be broken.
I guess it should be better to add another class: TAESCTRNIST for instance, with your patch.
What do you think?

synopse: note: using 64-bit only is for obvious security reasons, if the IV is not random but forged. 
So NIST algorithm is unsafe if the IV can be forged.

synopse:
No feedback. Closing.

SynCrypto.pas
karwos: Wikipedia is Wikipedia, and NIST guidelines are NIST guidelines.

The IV for CTR is **always** 128 bits. It is always AES operation on 128bit data, which result is XORed with plaintext (or ciphertext for decryption).
Somewhat common standard is **IV = NONCE + CTR**
DIffrence is how it is constructed - which bits are static one, which are random one.
The random one is called "nonce" and should be unique for each message (this is critical part), and static part (usually 00h 00h 00h...) called "counter".

Most crypto systems was breaken not because crypto systems itself failed, but because weak implementation (and weak knowledge of crypto systems). 

Because SynCrypto expects TAESIV (16 bytes array) in TAESCTR, not dedicated struct with 64-bit nonce value, it is entirely programmer choice to follow standards and carefully implement cryptography. This is weak point here.

I don't understand and don't agree with You that NIST algorithm is unsafe. Any CTR algorithm is unsafe if IV value (thus one created with NONCE+COUNTER approach aswell) is reused, which will lead to XOR stream recovery (known plaintext attack).
There is completly no diffrence if You reuse 64-bit nonce (and mix with 00h 00h 00h counter) or reuse 128-bit IV as a whole. For example, when generated with weak PRNG (or no PRNG at all; like GetTickCount which is very predictable and already seen such implementations).

I understand that my patch would break everything - especially already encrypted data with SynCrypto would lead to corruption, which is unacceptable.

Back to the NIST. NIST documents even don't even specify the counter length. It may be less than 64 bits, but if you use lower value, and counter will roll, you are screwed. So 64 bit is feasible value (2^64 * 16 bytes). This is understandable, because NIST guidelines are for generic symmetric block ciphers, which varies in block size (3DES, i.e.).

According to NIST B.2:

> For example, let the standard incrementing function apply to the five **least significant bits** of
> eight bit blocks, so that b=8 and m=5 (unrealistically small values); let * represent each unknown
> bit in this example, and let ***11110 represent a block to be incremented. The following
> sequence of blocks results from four applications of the standard incrementing function:
> * * * 1 1 1 1 0 
> * * * 1 1 1 1 1 
> * * * 0 0 0 0 0 
> * * * 0 0 0 0 1 
> * * * 0 0 0 1 0. 

In our case this is b=64 and m=64

And NIST B.3:

> In the first approach**, for a given key, all plaintext messages are encrypted sequentially**. Within
> the messages, the same fixed set of m bits of the counter block is incremented by the standard
> incrementing function. **The initial counter block for the initial plaintext message may be any
> string of b bits.** The initial counter block for any subsequent message can be obtained by
> applying the standard incrementing function to the fixed set of m bits of the final counter block
> of the previous message. In effect, all of the plaintext messages that are ever encrypted under the
> given key are concatenated into a single message; consequently, the total number of plaintext
> blocks must not exceed 2m . Procedures should be established to ensure the maintenance of the
> state of the final counter block of the latest encrypted message, and to ensure the proper
> sequencing of the messages.

> A second approach to satisfying the uniqueness property **across messages** is to assign to each
> message a unique string of b/2 bits (rounding up, if b is odd), in other words, a message nonce,
> and to incorporate the message nonce into every counter block for the message. The leading b/2
> bits (rounding up, if b is odd) of each counter block would be the message nonce, and the
> standard incrementing function would be applied to the remaining m bits to provide an index to
> the counter blocks for the message. Thus, if N is the message nonce for a given message, then
> the jth counter block is given by Tj
>  = N | [j] , for j = 1…n. The number of blocks, n, in any
> message must satisfy n < 2
> m . A procedure should be established to ensure the uniqueness of the
> message nonces. 

So as You see, both approaches are NIST followed. Your current implementation is actually second NIST approach, expect part of LSB incrementing requirement. This prones no security weaknesses, unless opposite 64-bit block is treated as nonce and changed each message encrypted. But I think that should be reflected into object parameters. Now it is passed as 16-bytes block, so programmer may assume that it is first NIST approach. That's why I've ran into problems trying to decrypt standarized cryptogram.

The first approach is also secure if IV is chosen uniquely across all message blocks/messages.

I know request is closed (but i saw already TAESCTRNIST implemented), but I think better approach would be to rebuild (with keep backward compatibility) current TAESCTR to give more control to user how actual IV block is constructed and follow NIST nomenclature/guidelines.

My proposition (i will open another pull request) is to add CompositeIV function with parameters BNonce, MCounter, and LSBCounter (for compatibility). Remember that some CTR implementations use 80 bit nonce + 48bit counter, also there is more weak implementation for 96 bits nonce + 32 bit counter.

This **all** follows NIST guidelines, but programmer should always remember to not encrypt more bytes than (2^counter*16), and select nonce with PRNG carefully. 

And the only reason I've opened that ticked I've ran into compatibility problems when dealing with cryptograms created with other NIST-following (but diffrent approach) rules. 

CTR is most vulnerable for bad implementation (when lack applied cryptography skills), no matter which approach You follow.

PR #291, opened 2020-03-21
new property TSQLDBConnectionProperties.ExecuteWhenConnected

pavelmash:
Array of statement to run on each connection just after connect
The usage scenarios examples:
   - Oracle: force case-insensitive like
    ['ALTER SESSION SET NLS_COMP=LINGUISTIC', 'ALTER SESSION SET NLS_SORT=BINARY_CI']
  -  Postgres: disable notices and warnings ['SET client_min_messages to ERROR']


synopse:
It was a very good and simple idea!

PR #293, opened 2020-03-22
fix headers regression

pavelmash:
Introduced in [c5224e46f0d30c]
My app use Content-length and Content-type: application/jon* headers from InHead string ( I pass InHead into JavaScript and and any of NPM lib what works with headers can read it)


synopse:
I have introduced a new HeadersUnFiltered optional parameter to THttpServer.Create - as an alternative to this modification.

Our RESTful stack (ORM + SOA) doesn't need this information at all, since by convention ContentType='' is identified as JSON, and all connection specific headers are available outside of the headers. It slightly reduce memory usage, and speed up headers lookup.

This new flag would please your specific requirement, I hope.

PR #294, opened 2020-03-24
mark TSynTestsLogged overloaded constructor as virtual

pavelmash:
- it overrided in TTestDatabaseBenchmark class
without this changes my FPC linux not compile new DB bench project


synopse:
I forgot to commit a file.
Sorry.

PR #295, opened 2020-03-27
Fix armfakestub when compiling with -O2

LongDirtyAnimAlf:
When compiling on ARM with -O2, some test do fail with strange errors.
I have looked into this issue. And found the cause !

There are not that many registers on ARM. In theory, R0-R3 are volatile, R4 and other could be freely used. So, we use register R4 to transport the method index into the fakecallstub.
However.
The optimizer does not know (and cannot know) that we are spoiling R4. So, during -O2 the compiler uses R4 without safeguarding it during jumps. Especially during the FakeCall.

There is only a single scratch-register on ARM that is not used by the compiler : IP/R12.
But that register is used as a temporary store for the stackpointer inside a stackframe generation.
So, I had to revert to pure assembler to solve this.

This pull works very well under all conditions. And it is guaranteed to work due to the use of the only register that can always be spoiled !
This pulls is the best thing we can do for all users: it will secure the function of the FakeCall.

Thanks, Alfred.
Ps: we should perhaps do the same for ARM64 ... ;-)


LongDirtyAnimAlf:
I have left the original code inside the source.
This is important when the pure assembler needs to be changed.
It is needed to design the pure assembler FakeStub.

PR #297, opened 2020-03-28
Prevent extended errors on ARM.

LongDirtyAnimAlf:
The latest commit for floats and more causes some errors on ARM systems.
These errors can be prevented by disabling (undefining) some features for ARM:
{$define EXTENDEDTOSHORT_USESTR} // this must be disabled for ARM(64)

However, this resulted in compilation errors.
This commit resolves these compilation errors and gives all greens on ARM(64).
(if EXTENDEDTOSHORT_USESTR is not enabled)
(this is not included in this pull: awaiting approval)

For the time being this solves.
For the long term, something else has to be done.


synopse:
Could you please elaborate which errors occur?

Could you try https://synopse.info/fossil/info/4a50dee9e8 ?

LongDirtyAnimAlf:
Looks 100% good now ! Thanks again.

mORMotFloatARM64

PR #301, opened 2020-04-03
fix load libpq.dll under Windows

pavelmash:
in case SynDBPostgresLibrary contains a full path to libpq.dll cwd should be changed for success library load (taken from @ab patch to TZNativeLibraryLoader.ZLoadLibrary)

Verified under Win10 with x64 libpq.dll 11.7.2 & 12.2 - both works


synopse:
Please check https://github.com/synopse/mORMot/commit/c62e6bbfad5c544b130ea73ad5d96e0a2e895410

PR #311, opened 2020-05-15
* Missing DocVariantType.Free

HgAlexx:
Hello,

Without this free, there is sometimes an Access Violation exception in function ClearVariantTypeList of unit System.Variants which try to free this object which memory seams to be freed/corrupt.

ClearVariantTypeList loop the variable "LVarTypes: array of TCustomVariantType;" and call .Free if not nil.

Adding the DocVariantType.Free into finalization section of unit SynCommons calls TCustomVariantType.Destroy which correctly free the object and removes it from LVarTypes, removing the access violation.

Regards,

### DIFF
--- a/SynCommons.pas
+++ b/SynCommons.pas
@@ -62772,6 +62772,9 @@ initialization
 
 
 finalization
+  {$ifndef NOVARIANTS}
+  DocVariantType.Free;
+  {$endif NOVARIANTS}
   GarbageCollectorFree;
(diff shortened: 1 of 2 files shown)


HgAlexx:
The second commit is more of a feature request.

The way we use THttpServerApi, we need access to the full url, this second commit add property FullURL on THttpServerRequest and populate it with value from Req^.CookedUrl.pFullUrl

Regards,

PR #314, opened 2020-05-20
Windows compilation fixed for SynCurl

pavelmash:


synopse:
I guess https://github.com/synopse/mORMot/commit/d9a6591f9ecd74a36374cd978daab4ec1a75dcc3 fixes this - with no dependency to SynCommons (not needed here).

PR #315, opened 2020-05-28
Read / Write blob column to / from a TStream

algalg:
I have implement chunked reading and writing of blobs from and to a TStream.  Can you please include this in the master repository?

### DIFF
--- a/SynDB.pas
+++ b/SynDB.pas
@@ -419,6 +419,11 @@ type
     function ColumnBlob(Col: integer): RawByteString; overload;
     /// return a Column as a blob value of the current Row, first Col is 0
     function ColumnBlobBytes(Col: integer): TBytes; overload;
+
+    procedure ColumnBlobToStream(Col: integer; Stream: TStream); overload;
+    procedure ColumnBlobWriteFromStream(Col: integer; Stream: TStream); overload;
+
+
     /// return a Column as a TSQLVar value, first Col is 0
     // - the specified Temp variable will be used for temporary storage of
     // svtUTF8/svtBlob values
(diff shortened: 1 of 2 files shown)


algalg:
Hello, 

I have implemented the ColumnBlobToStream and ColumnBlobWriteFromStream methods in TSQLDBStatement class and improved the code comments.

synopse:
I guess your comment about ColumnBlobWriteFromStream() is pretty incorrect.
I don't understand what it is supposed to do in practice...

synopse:
Your default ColumnBlobWriteFromStream() was buggy and triggered a GPF - WriteBuffer(pointer(BlobText)^,...) was needed.

algalg:
The comment about ColumnBlobWriteFromStream() is actually correct. 

It writes the data from a TStream into the BlobColumn in the database. 

SQLDBRows := FConnection.Execute('select a from T WHERE r_object_id = 123456 **for update**', []); // **for update** to get the lock on the row to save the blob to

SQLDBRows.Step();

FileStrm := TFileStream.Create('c:\test.txt', fmOpenRead or fmShareDenyNone);
try
  SQLDBRows.ColumnBlobWriteFromStream(0, FileStrm); // Write Blob to database
finally
  FileStrm.Free;
end;
SQLDBRows := nil;

synopse:
It is too Oracle specific, I am afraid.

algalg:
My intention is to also implement ColumnBlobWriteFromStream in the SynOleDB. 

I had implementation of BindBlobStream()  method using a temporary blob on oracle site, but the performance was much worse then writing to the blob column directly.

PR #324, opened 2020-06-20
Feature/systemd master

pavelmash:
Systemd integration:
 - **adedd** export required libsystemd functions (new unit `SynSystemd.pas`)
 - **added** THTTPServer now able to use externally created listening socket
 - **added** SynLog can echo directly into journald (with log level support) - new property `TSynLogFamily.EchoToConsoleUseJournal`
 - **added** samples 36 is extended to show a systemd socket activation (see SQLite3/Samples/36 - Simple REST Benchmark/README.md)
 - **added** LogView understand some kind of journald export (limited for a while by singe PID)
 - **fixed** LogView program can be compiled under Linux
 - **fixed** `createSuspended` parameter added into THttpServer constructor (in real app server should be created in suspended state to allow configure it properties before accepting client requests)
- I mark a TCrtSock.SockSendFlush as virtual to allow override it in descendants. For example it's allow me to create a "fake" socket and push it to `ThreadPool.Push(fakeSocket)`

### DIFF
--- a/SQLite3/Samples/11 - Exception logging/LogViewMain.lfm
+++ b/SQLite3/Samples/11 - Exception logging/LogViewMain.lfm
@@ -14,7 +14,7 @@ object MainLogView: TMainLogView
   OnCreate = FormCreate
   OnKeyDown = FormKeyDown
   OnShow = FormShow
-  LCLVersion = '1.7'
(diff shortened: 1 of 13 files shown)


pavelmash:
Yes, SynFPCLinux is still small, so as you wish

PR #328, opened 2020-06-26
**CRITICAL** ExistNameValue should return nil 

pavelmash:
ExistNameValue should return nil in case name not found (as expected by callers) but instead returns #0.
Issue introduced by [bde936aecd]
Without fix GetHeaderValue returns random result in case header does not exists

### DIFF
--- a/SynCrtSock.pas
+++ b/SynCrtSock.pas
@@ -3724,7 +3724,7 @@ begin
       inc(result);
     while result^<=#13 do
       if result^=#0 then
-        exit else
+        exit(nil) else
         inc(result);
   until false;
 end;


pavelmash:
@synopse - so that I no longer do this: what's wrong with 
```
exit(nil)
``` compared to
```
result := nil; exit;
``
?

synopse:
Exit(Nil) doesn't compile on older Delphi. :)Le 26 juin 2020 18:49, pavelmash <notifications@github.com> a écrit :
@synopse - so that I no longer do this: what's wrong with
exit(nil)
``` compared to

result := nil; exit;
``
?

—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

PR #329, opened 2020-07-01
use TMemo instead of TMemoEx for FPC on all platforms

pavelmash:


pavelmash:
Closed in flavor of [c231bf91297287]

PR #334, opened 2020-07-27
Socket based HTTP server 10 second delay for req with empty body

pavelmash:
If neither Content-Length nor Transfer-Encoding: chunked header present we should not try to read HTTP body - see https://greenbytes.de/tech/webdav/rfc7230.html#message.body.length

In other case we got a 10 second delay in case on non keep-alive requests:
```
curl -v 'http://localhost:8881/root/timestamp' -X OPTIONS --compressed -H 'Access-Control-Request-Method: POST' -H 'Access-Control-Request-Headers: authorization,content-type' -H 'Referer: http://localhost:6060/' -H 'Origin: http://localhost:8881' -H 'DNT: 1' -H 'Connection: Close'
```

### DIFF
--- a/SynCrtSock.pas
+++ b/SynCrtSock.pas
@@ -6863,18 +6863,10 @@ begin
     SetLength(Content,ContentLength); // not chuncked: direct read
     SockInRead(pointer(Content),ContentLength); // works with SockIn=nil or not
   end else
-  if ContentLength<0 then begin // ContentLength=-1 if no Content-Length
-    // no Content-Length nor Chunked header -> read until eof()
-    if SockIn<>nil then
-      while not eof(SockIn^) do begin
-        readln(SockIn^,Line);
-        if Content='' then
-          Content := Line else
-          Content := Content+#13#10+Line;
-      end;
-    ContentLength := length(Content); // update Content-Length
-    exit;
-  end;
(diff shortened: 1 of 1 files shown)


pavelmash:
According to [rfc7230 3.3.3 Message body length](https://greenbytes.de/tech/webdav/rfc7230.html#message.body.length) body exists if either  `Content-Length` or `Transfer-Encoding: chunked` specified. IMHO fix is correct from HTTP POW. May be existed expectations/behaviors is wrong?

My real use case is Microsoft library for WebDav Client (my server implements WebDav protocol to allow to open/edit/save Office files directly from server) - on some Windows it sends OPTION request with `Connection: Close' and without `Content-Length: 0`, so my server wait 10 second for every response

synopse:
The original code was trying to implement the following point in your link: 
_7. Otherwise, this is a response message without a declared message body length, so the message body length is determined by the number of octets received prior to the server closing the connection._

What is the HTTP status code from the WebDav client response in your case?
Perhaps we may check this.

Or just avoid reading the content for OPTION or HEAD verbs...

synopse:
Please try my latest fix https://github.com/synopse/mORMot/commit/ceebf11b26d17f549cf8b9c413a704e3eebd4447

PR #336, opened 2020-08-01
fixed FindFiles sub-folders search under Windows

eugeneilyin:
```Delphi
(F.Attr and (faDirectory {$ifdef MSWINDOWS}and faHidden{$endif})=faDirectory)
```
Always return `False` (due to `faDirectory and faHidden = 0` under Windows).
This bug is preventing any search in subfolders.


synopse:
Funny enough, it was fine in mORMot2 fork. ;)

PR #338, opened 2020-08-05
new property TSQLDBStatement.SQLCurrent

pavelmash:
Getter returns an SQL statement in case prepare not called / not needed for the current statement
We loose it while moves SQLPrepared property from ISQLDBStatement into TSQLDBStatement


pavelmash:
Added new property TSQLDBStatement.SQLCurrent

PR #343, opened 2020-08-13
Loop improvement in function IsEqual(const A,B; count: PtrInt): boolean;

lguzzon:
Exit when result is false

### DIFF
--- a/SynCommons.pas
+++ b/SynCommons.pas
@@ -23007,7 +23007,7 @@ function IsEqual(const A,B; count: PtrInt): boolean;
 var perbyte: boolean;
 begin
   result := true;
-  while count>0 do begin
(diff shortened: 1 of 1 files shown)


synopse:
The point of the code is to be constant in time, as expected by best cryptographic practices.
Your modification make it slightly faster, but not constant in time.
The SynCommons.pas function CompareMem() is to be used for fast and non-constant binary comparison.

lguzzon:
> The point of the code is to be constant in time, as expected by best cryptographic practices.
> Your modification make it slightly faster, but not constant in time.
> The SynCommons.pas function CompareMem() is to be used for fast and non-constant binary comparison.

Ok,
thanks for your time

PR #345, opened 2020-08-21
log view for files w/o header

pavelmash:
LogView improvements:
 - decrease thread panel width to 200px (maximize log space)
 - recognize regular TSynLog file without header (header can be removed for security reason)


synopse (approved):
Great!

PR #346, opened 2020-08-24
RemoveCommentsFromJSON will also remove a trailing commas

pavelmash:
Some JSON parsers (like mORMot) allows [trailing commas](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas), some not (Browsers).
But it's very conveniently in config files, for example to comment a last array element etc.
```
{
  "someArray": [
     "val",
     // "val2"
  ]
}
```

Also prevent reading of a chracter out from a string for "" case


synopse (approved):
good idea!

PR #349, opened 2020-09-02
RemoveCommentsFromJSON will remove trailing comma followed by comment

pavelmash:
for example comma after "Socket"
```
{
  "httpServer": {
    "host": "*",
    "port": "8881",
    "serverType": "Socket",
    /*"reverseProxy": {
      "kind": "nginx",
      "sendFileLocationRoot": "snake-ukrpatent-local"
    }*/
  },

```

### DIFF
--- a/SynCommons.pas
+++ b/SynCommons.pas
@@ -57834,7 +57834,38 @@ begin
 end;
 
 procedure RemoveCommentsFromJSON(P: PUTF8Char);
-var PComma: PUTF8Char;
+var  PComma: PUTF8Char;
+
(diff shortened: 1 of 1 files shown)


pavelmash:
@synopse, please, merge this.
Changes just looks big, but In fact here we just move a part of code (without changes) into inner function tryRemoveComment, because it needed to be called twice.
I apply this patch to my productions 3 month ago and it works in the wild without a problem

synopse (approved):
We could make it a bit faster by returning the P pointer and not parsing the ' ' again.

PR #359, opened 2020-10-09
SQL Server ODBC: set length for varchar parameters

pavelmash:
Some statements do not accept nvarchar(max), for example Full Text Search CONTAINS(field, ?)
Patch verified under Linux with UnityBase regression tests


synopse (approved):
Nice catch!

PR #360, opened 2020-10-28
fix hangs of sqlite3 prepare for statement like '/**'

pavelmash:
Due to error in my App I try to execute a statement `/**`
This hangs a prepare()
Not sure using of SQLITE_EMPTY is correct here


synopse (approved):
Nice!

PR #364, opened 2020-11-04
Update SynWinSock.pas

danielkuettner:
important in case of using self signed certs to prevent SEC_E_UNTRUSTED_ROOT result in HandshakeLoop


synopse (approved):

synopse:
make sense!
Thanks for the feedback!

PR #369, opened 2020-12-11
SynDBOracle - fix removing of ending semicolon

pavelmash:
fix removing of ending semicolon for statements ends with END what is part of an identifier, like
```
ALTER TABLE BLA_BLA DROP COLUMN STR_PREPEND;
```
or
```
SELECT SOMETHING FROM PERIODICALEND;
```


synopse (approved):
Tricky but if it works....

PR #371, opened 2020-12-19
Update mORMotHttpServer.pas

devaex:
Avoid Excpetion on finalization of HttpServer.
Lazarus + FPC , Win64.

Sample project to reproduce:
https://pastebin.com/uZFyhHi8

Forum thread:
https://synopse.info/forum/viewtopic.php?id=5691

### DIFF
--- a/SQLite3/mORMotHttpServer.pas
+++ b/SQLite3/mORMotHttpServer.pas
@@ -55,8 +55,8 @@ unit mORMotHttpServer;
  - can be called by any JSON-aware AJAX application
  - can optionaly compress the returned data to optimize Internet bandwidth
  - speed is very high: more than 20MB/sec R/W localy on a 1.8GHz Sempron,
-   i.e. 400Mb/sec of duplex raw IP data, with about 200 µs only elapsed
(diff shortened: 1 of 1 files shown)


synopse:
I am not sure this fix is the best way to circumvent the problem.
We need to find the root cause first - the TSynLog problem is not the root cause AFAIK.
This week, I was sick so AFK, but I will try to investigate further next week.

devaex:
Hi ab,

I hope you are well.

I agree with your observation. I tried to find the source of the problem but I couldn't. I sent this pull request, because I believe that removing DBServers in destroy does not break the code and would be good practice, or not?

Take your time, the important thing is to take care of your health.

I wish you well.

devaex:
Some findings:

- As pointed by pvn occurs with httpApi only.

- It is not directly related to the logs, any routine that adds a small delay to EndCurrentThread, such as a writeln will cause the error.

- Maybe it's a race condition problem.

It is difficult to debug. Because the debug itself can prevent the race condition from occurring.

pavelmash:
Problem is solved by [caeb7bdf1] and finally by [3ab2c8860].

PR #374, opened 2021-01-18
Initialize Gdip object to prevent AV when NOTSYNPICTUREREGISTER is set

AntoineGS:


synopse:
Gdip is already equal to nil by default at startup.

AntoineGS:
Ah right, I mis-diagnosed the issue.
I am working on a project that loads the mormot files dynically through a .bpl and on the second unload it gives an AV on the Gdip.Free.
Delphi (at least 2010) is not re-initializing Gdip to nil on the second LoadPackage.
If I re-commit the FreeAndNil(Gdip) is it ok?

PR #377, opened 2021-02-03
Update SynDBFireDAC.pas

juanter:
Init with LoginPrompt=false


synopse (approved):
It may change the current behavior, but this behavior is indeed unwanted, especially for a service!

PR #378, opened 2021-02-07
added public access to the TSynThreadPool worker threads

pavelmash:
Some program can need a direct access to the thread pool workers, for example I use it to terminate a blocking operations inside workers under Linux:
```
 // send an INT signal to all workers to break a possible blocking opeartions
 for i := 0 to myHttpServer.ThreadPool.WorkerThreadsCount - 1 do
   pthread_kill(myHttpServer.ThreadPool.WorkerThreads[i].ThreadID, SIGINT); 
```

### DIFF
--- a/SynCrtSock.pas
+++ b/SynCrtSock.pas
@@ -855,6 +855,7 @@ type
     function PopPendingContext: pointer;
     function QueueLength: integer; virtual;
     {$endif USE_WINIOCP}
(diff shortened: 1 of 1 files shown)


synopse:
Has been included as https://github.com/synopse/mORMot2/commit/18bb8e312a6815e467d53a2f340c57d80bd7db2c

PR #379, opened 2021-02-08
fix: it brings ftGuid as string

mdbs99:
I have a system that uses GUID fields a lot, and DataSetToJSON() always returns "null" on these fields.
This PR fix the translation from DataSet to JSON.


synopse (approved):
Thanks for sharing!

PR #384, opened 2021-03-05
Support for Batch operations in UniDAC

matkov:
1. Support for batch operations in the UniDAC is added
2. Specific option 'UseUnicode' for SQLite is set to True to avoid broken chars that are not from current locale
3. Do not set specific option 'UseUnicode' for FPC because UniDAC returns all strings in UTF8 encoding for FPC (to avoid convertation UTF8 -> UTF16 -> UTF8)
4. Calling the Prepare method is added for TUniQuery that improves performance
5. Fix metadata returning for Oracle provider

### DIFF
--- a/SynDBDataset/SynDBUniDAC.pas
+++ b/SynDBDataset/SynDBUniDAC.pas
@@ -175,6 +175,7 @@ type
   ///	implements a statement via a UniDAC connection
   TSQLDBUniDACStatement = class(TSQLDBDatasetStatement)
   protected
+    fBatchExecute: Boolean;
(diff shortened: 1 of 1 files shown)


synopse (approved):
Nice!

synopse:
One question: Will it work with any version of UniDAC?

matkov:
> 
> 
> One question: Will it work with any version of UniDAC?

Support for Batch operations was added in 2015 year (version 6.2.7). I think incompatibility with versions before 6.2.7 is not a problem.

synopse:
Could you please check https://synopse.info/forum/viewtopic.php?id=5816 ?

PR #387, opened 2021-03-16
added optional result size limit for ISQLDBRow.FetchAll..

pavelmash:
throw if limit exceed - this is better when OutOfMemory for example


synopse:
I preferred to put the parameter in TSQLDBConnectionProperties. With a default value of 512MB.
And use it for JSON, Binary and CSV result sets.
I will port it to mORMot 2 also.
Thanks for the idea: I always wanted to add this, and always forgot to implement it.

PR #389, opened 2021-04-08
Added dPostgreSQL to DB_HANDLECREATEINDEXIFNOTEXISTS.  PostgreSQL supports CREATE INDEX IF NOT EXISTS. See https://www.postgresql.org/docs/9.5/sql-createindex.html

wxinix:


synopse (approved):
Fine!

PR #390, opened 2021-04-11
Fixed TRawUTF8List.SetText.  The existing code didn't refresh line count after setting text.  As a result,  TRawUTF8List.Count would erroneously return 0, even SetText has set the lines.

wxinix:


synopse (commented):
It doesn't make any sense.
fValules.Count holds a pointer to fCount.

synopse:
Please send a reproducible example and use the forum for discussion.

wxinix:
You are right. Sorry for the false alarm.  It turns out to be something 
else.  Sorry again.

------ Original Message ------
From: "ab" ***@***.***>
To: "synopse/mORMot" ***@***.***>
Cc: "Wuping Xin" ***@***.***>; "Author" 
***@***.***>
Sent: 4/12/2021 3:10:22 AM
Subject: Re: [synopse/mORMot] Fixed TRawUTF8List.SetText. The existing 
code didn't refresh line count after setting text. As a result, 
TRawUTF8List.Count would erroneously return 0, even SetText has set the 
lines. (#390)

>
***@***.*** commented on this pull request.

PR #395, opened 2021-05-01
TODBCConnectionProperties.SQLStatementTimeoutSec + ODNB doubles binds

pavelmash:
- introduce TODBCConnectionProperties.SQLStatementTimeoutSec
- fix SQL Server binding for doubles in range (-1; 1)


synopse (approved):
Seems just fine!

PR #401, opened 2021-08-13
use storedName from original ZipEntry instead ob zipName

pavelmash:
`zipName` contains path with `/` replaced to `\` (by `TZipRead.Create`), but acoding to ZIP file format all slashes MUST be `/`.
The same issue is in mORMot2


synopse (approved):
This may be enough for mORMot 1.
On mORMot 2, I will make an option as DotNet 4.6.1. does.
https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/mitigation-ziparchiveentry-fullname-path-separator

PR #404, opened 2021-09-13
set a curl timeout as SendTimeout+ReceiveTimeout

pavelmash:
set a curl timeout as SendTimeout+ReceiveTimeout - this prevent request to wait forever;
 use *_TIMEOUT_MS (available in CURL since 2007year)

### DIFF
--- a/SynCrtSock.pas
+++ b/SynCrtSock.pas
@@ -11919,11 +11919,9 @@ begin
   fHandle := curl.easy_init;
   if curl.globalShare <> nil then
     curl.easy_setopt(fHandle,coShare,curl.globalShare);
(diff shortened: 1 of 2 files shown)


synopse (approved):
Nice!

synopse:
My guess is that the correct logic may be
```   if SendTimeout < ReceiveTimeout then
    SendTimeout := ReceiveTimeout;
  if SendTimeout <> 0 then // prevent send+receive forever
    curl.easy_setopt(fHandle, coTimeoutMs, SendTimeout);
```

pavelmash:
Ok, let it be as  you propose `max(SendTimeout, ReceiveTimeout)` - I will investigate how it works in  productions

PR #405, opened 2021-09-16
Implementation of ColumnBlobToStream and ColumnBlobFromStream in TOleDBStatement

algalg:
Implementation of ColumnBlobToStream and ColumnBlobFromStream in TOleDBStatement
Implementation of TransactionIsolationLevel property in TOleDBConnectionProperties

### DIFF
--- a/SynDB.pas
+++ b/SynDB.pas
@@ -719,7 +719,11 @@ type
     // - parameters marked as ? should have been already bound with Bind*() functions
     // - should raise an Exception on any error
     // - after execution, you can access any returned data via ISQLDBRows methods
-    procedure ExecutePrepared;
+    procedure ExecutePrepared; overload;
+    /// execute a prepared SQL statement
+    // - if ForUpdate - true then expected to be used with 'SELECT' locking statements.
+    // - will prepare statement for ColumnBlobFromStream use
+    procedure ExecutePrepared(ForUpdate: boolean); overload;
     // execute a prepared SQL statement and return all rows content as a JSON string
     // - JSON data is retrieved with UTF-8 encoding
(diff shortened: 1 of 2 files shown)


synopse (changes requested):
I am worried about the ExecutePrepared() overload change.
I am not sure it would work as such with the ORM, and it may definitively break the other DB providers, outside OleDB.
My guess is that we could guess which kind of statement is to be prepared from the SQL.

algalg:
I use this on MSSQL Server. Unfortunately I don't think there is any way to tell  that the select query is for updating the content. 
This odd construction is what MS documents as way of working with large data https://docs.microsoft.com/en-us/sql/connect/oledb/ole-db-how-to/set-large-data-ole-db?view=sql-server-ver15

This is how it is used in the code:
`
  objectid := '123';
  FileName := 'C:\TheBigContentFile.pdf';
  
  FileStrm := TFileStream.Create(FileName, fmOpenRead or fmShareDenyNone);
  try
    FProviderSession.ExecuteNoResult('INSERT INTO CONTENT_TABLE (id, blobfiled) ', [objectid, 'NULL']);

    Stmt :=  Connection.NewStatementPrepared('SELECT blobfiled FROM CONTENT_TABLE WHERE id = ?', true, true);
    Stmt.ForceBlobAsNull := false;
    Stmt.Bind([ObjectID]);
    Stmt.ExecutePrepared(true);
    SQLDBRows := Stmt;

    SQLDBRows.Step();
    SQLDBRows.ColumnBlobFromStream(0, FileStrm);
  finally
    FileStrm.Free;
  end;`

synopse:
No other proposal after a few months waiting for an alternative.

Please reopen if needed.

PR #414, opened 2021-11-16
force OS in Zip header to be Unix

pavelmash:
- in this case all zip clients correctly recognize Unicode file names. For 0 (DOS) - most GUI clients ignore UTF8 flag

### DIFF
--- a/SynZip.pas
+++ b/SynZip.pas
@@ -390,7 +390,7 @@ type
   TFileHeader = {$ifdef USERECORDWITHMETHODS}record
     {$else}object{$endif}
     signature     : dword;           // $02014b50 PK#1#2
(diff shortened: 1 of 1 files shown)


pavelmash:
See https://synopse.info/forum/viewtopic.php?id=6052 forum thread for details

pavelmash:
[utf8FileNamesAfterFix.zip](https://github.com/synopse/mORMot/files/7549008/utf8FileNamesAfterFix.zip)
[utf8FileNamesBeforeFix.zip](https://github.com/synopse/mORMot/files/7549009/utf8FileNamesBeforeFix.zip)

Two files (before fix and after fix) contains cyrillic file names attached

synopse (approved):

PR #415, opened 2021-11-30
ODBC: do not throw on unknown driver

pavelmash:
This allows to access any database via ODBC even if mORMot does not know it


synopse:
My guess is that a flag to ignore the name, and update the internal list (once a driver is tested) is better.

That is, keep the default behavior for safety, and only if you use an "unsupported driver", just set the flag and enjoy.
Of course, adding the unsupported driver names to the internal list is the best solution.

synopse:
Closed since https://github.com/synopse/mORMot2/commit/a5ab3c3acf2bbcd2f416d5545fe2b654045a3e20 should do the trick.

Please report on the forum and/or reopen this pull request if needed.

PR #418, opened 2021-12-03
Bug in SecDecrypt

achechulin:
There is a bug in SecDecrypt, as reported in forum thread:
https://synopse.info/forum/viewtopic.php?pid=36311

Decryption is done in-place, so call to FreeContextBuffer is wrong.


synopse (approved):
Great!

PR #419, opened 2021-12-10
Fix so it compiles again under D2007

JavierusTk:
Now there si a new type TSSPIBuffer, but in a few places variables are declared as RawByteString instead of TSSPIBuffer; it works with Unicode Delphis, as they declare TSSPIBuffer = RawByteString, but in D2007 is different: TSSPIBuffer = AnsiString, so the compiler complains and fails


synopse:
I guess your fix would break the Linux GSSAPI support.

Please check https://github.com/synopse/mORMot/commit/56ccaaee198baa59e2314a256cfba4c170abbea6

synopse:
No news since answered - closing.

Please reopen if needed.

PR #423, opened 2021-12-22
Fix fill region bugs in SynPDF

iadcode:
These changes fix some bugs in SynPDF relating to filling regions.
- The first commit ensures complex regions are filled using the defined region rather than the bounding rectangle
- The second commit fixes a bug where the current brush is incorrectly changed with filling a region

Before the changes
image
After the changes
image

This behaviour can be tested with code such as
```
  procedure EmfToPdf(EmfFilename: String; PdfFilename: String);
  const
    C_PDFPPI = 72;
    C_MMperInch = 25.4;
  var
    PDFDocument: TPdfDocument;
    Metafile: TMetafile;
    Scale: Single;
  begin
    Metafile := TMetafile.Create;
    try
      Metafile.LoadFromFile(EmfFilename);
      PDFDocument := TPdfDocument.Create(False, 0, False);
      try
        PDFDocument.DefaultPaperSize := psA4;
        PDFDocument.ScreenLogPixels := C_PDFPPI;
        PDFDocument.AddPage;
        Scale := C_PDFPPI / Trunc((Metafile.Width * 100) / (Metafile.MMWidth / C_MMperInch));
        PDFDocument.Canvas.RenderMetaFile(Metafile, Scale, Scale, 0, 0, tpSetTextJustification, 99, 101, tcAlwaysClip);
        PDFDocument.SaveToFile(PdfFilename);
      finally
        FreeAndNil(PDFDocument);
      end;
    finally
      FreeAndNil(Metafile);
    end;
  end;
```
And using an EMF such as the one in this zip:
[BugsEMF.zip](https://github.com/synopse/mORMot/files/7759527/BugsEMF.zip)

### DIFF
--- a/SynPdf.pas
+++ b/SynPdf.pas
@@ -8954,6 +8954,12 @@ type
     width: Single;
   end;
 
+  TPdfEnumStateBrush = record
+    null: boolean;
+    color: integer;
+    style: integer;
+  end;
+
   /// a state of the EMF enumeration engine, for the PDF canvas
   // - used also for the SaveDC/RestoreDC stack
   TPdfEnumState = record
@@ -8973,11 +8979,7 @@ type
     // current selected pen
     pen: TPdfEnumStatePen;
     // current selected brush
-    brush: record
-      null: boolean;
-      color: integer;
-      style: integer;
-    end;
+    brush: TPdfEnumStateBrush;
     // current selected font
     font: record
       color: integer;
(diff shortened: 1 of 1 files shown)


synopse:
Update: Why is this not merged yet?

We have merged a lot of fixes from third parties, and it almost every time broke the code working for other users of the library.
PDF generation is a tricky matter, and it is easy to break the existing.

From our point of view:
1) Since we don't actively support mORMot 1.18 any more, you are welcome to switch to mORMot 2 and its https://github.com/synopse/mORMot2/blob/master/src/ui/mormot.ui.pdf.pas unit
2) Fixing for one EMF input is likely to break existing code, which works for most users, so we would rather be very conservative, and only accept pull request with demonstrated compatibility
3) You should also ensure that the code strictly follow mORMot formatting and types, and also that it compiles on both FPC and Delphi.

PR #424, opened 2021-12-22
Fix problem with horizontal scaling in SynPDF

iadcode:
This change fixes a bug in SynPDF where the horizontal scaling is incorrect.
This issue can be found in this ticket from 2014, https://synopse.info/fossil/tktview?name=f1fdab52ce

### DIFF
--- a/SynPdf.pas
+++ b/SynPdf.pas
@@ -10597,14 +10597,6 @@ begin
       end else
         Canvas.ShowText(pointer(tmp));
     Canvas.EndText;
-    case Positioning of
-    tpSetTextJustification:
-      if nspace>0 then
-        Canvas.SetWordSpace(0);
-    tpKerningFromAveragePosition:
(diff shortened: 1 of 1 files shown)


synopse:
Update: Why is this not merged yet?

We have merged a lot of fixes from third parties, and it almost every time broke the code working for other users of the library.
PDF generation is a tricky matter, and it is easy to break the existing.

From our point of view:
1) Since we don't actively support mORMot 1.18 any more, you are welcome to switch to mORMot 2 and its https://github.com/synopse/mORMot2/blob/master/src/ui/mormot.ui.pdf.pas unit
2) Fixing for one EMF input is likely to break existing code, which works for most users, so we would rather be very conservative, and only accept pull request with demonstrated compatibility
3) You should also ensure that the code strictly follow mORMot formatting and types, and also that it compiles on both FPC and Delphi.

PR #434, opened 2022-11-16
fix: SynBidirSock.pas raise access violation errors when ProcessMask()

ysair:
1. run Project31WinHTTPEchoServer.exe
2. execute this test code:
```pascal
var
  Client :  THttpClientWebSockets;
  protocol : TWebSocketProtocol;
  frame: TWebSocketFrame;
begin
  protocol := TWebSocketProtocolChat.Create('meow', 'whatever');
  Client := THttpClientWebSockets.WebSocketsConnect(
      'localhost','8888',protocol, nil, '', 'whatever');
  frame.opcode  :=  focText;
  frame.content :=  [];
  frame.payload :=  'test1234';
  Client.WebSockets.SendFrame(frame); //av error
  Client.free;
end;
```

### DIFF
--- a/SynBidirSock.pas
+++ b/SynBidirSock.pas
@@ -2894,7 +2894,7 @@ begin
   result := GetBytes(pointer(data),hdr.len32);
   if result then begin
     if hdr.mask<>0 then
-      ProcessMask(pointer(data),hdr.mask,hdr.len32);
+      ProcessMask(@data[1],hdr.mask,hdr.len32);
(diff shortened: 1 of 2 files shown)


synopse (commented):
This is a very specific use case for sure.
Usually, you don't generate the frames from a constant.
And your fix won't work with FPC.

synopse:
I have made two commits to fix the RemoteIP issue and properly document TWebSocketFrame.payload use.

PR #435, opened 2022-11-20
Added ISQLDBStatement.LastStepReturnedValue, without affecting all previously existing functions

edwinyzh:


synopse:
This introduces a SQLite3 specific information into abstract ISQLDBStatement, which is not a good idea.

synopse:
No reaction to my above comment from 1 year ago.
So I can't merge this.
Please react here and re-open the ticket if needed.

PR #437, opened 2022-11-21
Fix for setting ParenthesisAfter property value in TSynTableStatement…

ASiwon:
… class while parsing SQL with complex condition like:

select field from table where id=? and (column1<>? or column1 is null)

Before correction this query is not correctly parsed in the class TSynTableStatement - closing parenthesis is missing. This causes SQL execution error in my case in Oracle, but probably also in any other database engines.

### DIFF
--- a/SynSelfTests.pas
+++ b/SynSelfTests.pas
@@ -10891,6 +10891,23 @@ begin
   Check((length(Stmt.Select)=2)and(Stmt.Select[1].Field=0)and
     (Props.Fields.List[Stmt.Select[0].Field-1].Name='Data'));
   Check(Stmt.OrderByField=nil);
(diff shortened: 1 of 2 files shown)


synopse (approved):
Seems great!
Thanks!

synopse:
Note that I won't port it to mORMot 2 because it is very specific, and handles your workcase as you expect, but is likely to fail in a more general usecase, e.g. when fields are renamed.

ASiwon:
Ok. But it will be possible to support similar conditions in mORMot 2? Null values occurs very often in the Oracle database - empty string is always treated as null, which makes that conditions must be created in specific way to take into account the specificity of the database system.

PR #444, opened 2023-03-22
Added support for non-english file names in Zip archives

pavelmash:
This is called [Info-ZIP Unicode Path Extra Field](https://pkwarecachefly.net/webdocs/casestudies/APPNOTE.TXT).
We discover what build-in zip on latest Windows10 Home put such names there.

Will be good to be ported into mORMot2


synopse (approved):

synopse:
Do you have a sample file so that I could validate it on my side?

pavelmash:
Customer give me such file, but it huge. I will ask for smaller

synopse:
Works fine on mORMot 2 too.
Thanks for the .zip file.

PR #457, opened 2024-10-14
fix invalid pointer cast when IMAGE_FILE_LARGE_ADDRESS_AWARE enabled …

zedxxx:
…and first 2G RAM reserved (delphi, win32)


synopse (approved):
Sounds fine.

Note tha mORMot 2 is not affected by this.

Issue #458, opened 2024-12-01
PasXip.pas issue.


ijbranch:
Hi Team,
I recently started to play in 64 bit.  Normally I was 32 bit exclusive.
I ran up against an issue with paszip.pas, used in a couple of places in various 3rd Party libraries.
I kept geting an error at line 4341 or thereabouts:  file_ := INVALID_HANDLE_VALUE;
Tracing, I found variable file_ was defined at line around 3963 as a DWORD:  file_: dword; 
This is fine in 32bit but problematic in 64bit.
The solution was to change the declaration to:   file_: THandle;
Now paszip.pas builds without issue in both 32 & 64 bit.
HTH in some way..

Regards,
Ian

synopse:
Please try with the above commit, and close if it is OK for you.

I tried to fix all wrong occurrences of DWORD in this unit.

ijbranch:
I was still experiencing issues with 64bit.
Examining INVALID_HANDLE_VALUE at the error line I found it to be a rediculous value:  1811674407373709551615.
Doing some research I found that INVALID_HANDLE_VALUE seems to be generally assigned a value of -1.
I made this change in PasZip:

{$WARNINGS OFF}
{$Q-,R-} // Turn range checking and overflow checking off

{ $D-,L-}

{$I Synopse.inc}

interface

uses
{$IFDEF MSWINDOWS}
  Windows,
{$ELSE}
  Types,
{$ENDIF MSWINDOWS}
  SysUtils;

  //////////////////////////////////////////////////////////////////////
const       //    This const add on 09/12/2024.
  INVALID_HANDLE_VALUE = THandle(-1); // Add the constant here
  /////////////////////////////////////////////////////////////////////

type
{$IFDEF HASCODEPAGE}
  RawByteZip = RawByteString;
  TZipName = type AnsiString(437);
{$ELSE}
  RawByteZip = AnsiString;
  TZipName = AnsiString;
{$ENDIF HASCODEPAGE}

I will advise if I have any more issues.

synopse:
Please try with the above commit.

Using a function like in mORMot 2 seems a cleaner way.

ijbranch:
No problem.  Either way seems to work.  
I will let you know after some more testing.

zedxxx:
These variables must also be `THandle` types:
  - `file_, map: dword;` here https://github.com/synopse/mORMot/blob/master/PasZip.pas#L249
[...]

PR #459, opened 2024-12-10
#458: another fix for PasZip

zedxxx:


synopse:
TFileHandle is not needed here. THandle is already integer/dword for oldest Delphi.

zedxxx:
Thanks.

synopse:
Thanks for the feedback! :smile:

PR #460, opened 2025-03-03
Increased buffer in GetHeader due to token length

SteveNew:
In an OIDC auth scenario - we lost the tokens, since they exceeded the allocated 4096 buffer slightly - increasing it solved the issue.

### DIFF
--- a/SynCrtSock.pas
+++ b/SynCrtSock.pas
@@ -6973,7 +6973,7 @@ var s,c: SockString;
     i, len: PtrInt;
     err: integer;
     P: PAnsiChar;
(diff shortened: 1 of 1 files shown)


synopse (approved):
Seems fair enough.
I limited to 4KB per line because it sounded like a reasonnable value, and avoid any DoS attack with huge/invalid input.
But 8KB is fine too in that respect. ;)

synopse:
Thanks for the proposal.

I also ported it to the mormot 2 repository.

SteveNew:
Thanks Arnaud! That saved me from doing a mORMot2 PR 👏

PR #472, opened 2026-05-04
Oracle: fetch all rows even in case of warning

pavelmash:
Oracle: fetch all rows even in case of warning. For example: ORA-24347 Warning of a NULL column in an aggregate function. The same actual for mORMot2


synopse (approved):
thanks!

